PR #3, opened 2020-01-13
Fixed XF button typo

GreenBeard:


RTimothyEdwards:
Thanks, I just took care of this.  I am not yet used to handling pull requests from the mirror site, so let me know if it didn't happen properly, but as far as I know, I followed my instructions and it worked.

PR #12, opened 2020-05-22
Changes to the LEF writer

mithro:
* Strip trailing whitespace from file.
 * Use defines for specifying the floating point accuracy and indenting.
 * Change to using two space indenting and 6 after decimal point accuracy.


RTimothyEdwards:
Merged.

PR #13, opened 2020-05-23
Add Makefile target which strips all trailing whitespace.

mithro:


mithro:
@RTimothyEdwards - After this is merged you should run `make fix-trailing-whitespace` and then commit the result.

RTimothyEdwards:
Pulled, and the script was run and the result was pushed back.

PR #14, opened 2020-05-23
Make CVS value expansion work in git.

mithro:
A lot of the files in repository have the old `$Header: XXX $` and `$Id: XXX $` style tags. 

This pull request removes the expanded tags in the code and uses [git filters + attributes](https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes#_keyword_expansion) to expand them on checkout.

For this to work, a person needs to run `git config --local include.path ../.gitconfig` to enable this to work so `make setup-git` is added to the Makefile.

 * `$Id$` and `$Revision$` gets expanded to the `git describe --always --long` output
 * `$Date$` gets expanded to the date of the last value commit value.
 * `$Author$` gets expanded to the email address of the person who last committed to the file.
 * `$Header$` gets expanded to `<filename> <git describe> <last modification commit date> <last modification author address>`.


RTimothyEdwards:
Pulled, but I have not followed the instructions above.  When you say "a person needs to run", do you mean me? "../.gitconfig" implies that there is a ".gitconfig" outside of the repo?  Why?

mithro:
@RTimothyEdwards A person needs to run `make setup-git` after checking out for the first time.

The `git config --local include.path ../.gitconfig` means `--local` means relative to local git config file (which is `.git/config`). So `<basepath>/.git/../.gitconfig` which is the `.gitconfig` file at `<basepath>/.gitconfig` inside the repo.

PR #15, opened 2020-05-23
LEF writer: Fix indenting.

mithro:
I missed the LAYER value and the geometry was one level too high.

Previous;
```
    PORT
         LAYER li1 ;
      RECT 1.145000 1.075000 1.690000 1.275000 ;
      RECT 3.720000 1.075000 4.490000 1.275000 ;
         LAYER met1 ;
      RECT 1.105000 1.260000 1.395000 1.305000 ;
      RECT 3.765000 1.260000 4.055000 1.305000 ;
```

After;
```
    PORT
      LAYER li1 ;
        RECT 1.145000 1.075000 1.690000 1.275000 ;
        RECT 3.720000 1.075000 4.490000 1.275000 ;
      LAYER met1 ;
        RECT 1.105000 1.260000 1.395000 1.305000 ;
        RECT 3.765000 1.260000 4.055000 1.305000 ;
```

### DIFF
--- a/lef/lefWrite.c
+++ b/lef/lefWrite.c
@@ -48,6 +48,7 @@ static char rcsid[] __attribute__ ((unused)) = "$Header: /usr/cvsroot/magic-8.0/
 #define IN0 "  "
 #define IN1 "    "
 #define IN2 "      "
+#define IN3 "        "
 
 /* ---------------------------------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


RTimothyEdwards:
Pulled.

RTimothyEdwards:
@mithro :  I modified the "lef write" routine again.  Now it uses various settings from the tech "cifoutput" section to set the UNITS;  e.g., for SkyWater with "gridlimit 5" (minimum manufacturing grid 5nm), the UNITS is set to 200.  All the values in the file are then limited to a precision of 0.005 microns, or 5 nanometers.   This makes more sense than your combination of 6-decimal-place precision with UNITS 1000.  Also, the ANTENNAGATEAREA and ANTENNADIFFAREA should, like CUTAREA, be left as %f and not formatted according to the precision for distances, since they are not distances.

PR #21, opened 2020-06-27
Changed the end-of-line comment character from ';' to '$' when

mooredan:
formatting a SPICE netlist for NGSPICE.  This is the documented
character as described in the NGSPICE User Manual, section 2.2.4
End-of-line comments

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,7 @@ install-tcl-real: install-tcl-dirs
 clean:
 	for dir in ${MODULES} ${PROGRAMS} ${TECH} ${UNUSED_MODULES}; do \
 		(cd $$dir && ${MAKE} clean); done
-	${RM} *.tmp */*.tmp *.sav */*.sav *.log TAGS tags
+	${RM} *.tmp */*.tmp *.sav */*.sav *.log TAGS tags */Depend
 
 distclean:
 	touch defs.mak
--- a/ext2spice/ext2hier.c
+++ b/ext2spice/ext2hier.c
@@ -553,7 +553,7 @@ spcdevHierVisit(hc, dev, scale)
 	case DEV_FET:
 	    if (source == drain)
 	    {
-		if (esFormat == NGSPICE) fprintf(esSpiceF, "; ");
+		if (esFormat == NGSPICE) fprintf(esSpiceF, "$ ");
 		fprintf(esSpiceF, "** SOURCE/DRAIN TIED\n");
 	    }
 	    break;
@@ -561,7 +561,7 @@ spcdevHierVisit(hc, dev, scale)
 	default:
 	    if (gate == source)
 	    {
-		if (esFormat == NGSPICE) fprintf(esSpiceF, "; ");
+		if (esFormat == NGSPICE) fprintf(esSpiceF, "$ ");
 		fprintf(esSpiceF, "** SHORTED DEVICE\n");
 	    }
 	    break;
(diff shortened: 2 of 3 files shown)


RTimothyEdwards:
My version of the ngspice manual says that either ";" or "$ " (semicolon alone, or dollar sign followed by a space) are both valid in-line comment delimiters.

mooredan:
Current version of the ngspice manual (version 32, May2, 2020) does mention ";"  but in the context of the .control section where it means "continuation" 
http://ngspice.sourceforge.net/docs/ngspice-manual.pdf

RTimothyEdwards:
For the hierarchical node name fix, I have to review this very carefully, as normally it is intentional to produce a flattened namespace and not specifically invoke the hieararchy within ngspice.  Indeed, anything that connects between two cells should be forcing a port into the subcell, so there should never be a need to bypass the subcircuit I/O and reference something down in the hierarchy directly.  Doing so constitutes an error in the extraction that needs to be fixed (i.e., failure to recognize that an additional port was needed, and to add it).

RTimothyEdwards:
Well, okay, then.  I wonder who there is deciding to make (apparently unnecessary?) changes to the file syntax that break backwards compatibility with older versions of ngspice?

RTimothyEdwards:
Is it possible for you to put the changes to the in-line comments in their own pull request, as I would like to treat the hierarchy delimiter problem as a bug (with a different solution)?

mooredan:
Understood. Where I have seen the hierarchical node references pop up (other than in the capacitor test case that I provided) when higher level metal is placed over the top of standard cells.  When the standard cells are extracted/netlisted it isn't aware of anything above, because it's not there.  I think that what its trying to do is add additional capacitance to the node underneath.  This all probably accounted for in flat extraction/netlisting.

RTimothyEdwards:
Yes, but ext2spice is supposed to understand the hierarchy as specified in the parent cell to add more ports to the child cell as needed.  For example, if you put a buffer into a layout and then contact the midpoint with a wire, and extract that, you will find that the buffer has gained a port.  My suspicion is that this just isn't being done in the case of coupling caps, and so the reference looks like a hierarchical node when it is supposed to be connecting to a port of the subcell (which has not been created like it was supposed to).

mooredan:
Admittedly, I'm a novice at this whole "pull request" business, so I'm flying by the seat of my pants -- and thankful that its working at all. Sure, I'll format things however you desire; my goal is to not make things any harder on you as it is.    In this case, it's not perfectly clear to me what you want, should I have not included the "/* more processing */" comment.  It was just a cut-and-paste from the line above.

RTimothyEdwards:
Maybe the right solution is what you're doing, to avoid adding new ports to a subcell, because enough coupling could cause all standard cells to end up with hundreds of I/O ports.  But I just want to check it carefully, because I have fears of it breaking a lot of things if changed.

RTimothyEdwards:
I just want "fix hierarchical node names" to be put in a separate pull request so I can go ahead and merge the other stuff while I'm still pondering over whether or not the hierarchy fix has unintended consequences somewhere.

mooredan:
Oh, I see.  Actually I was surprised that github did this grouping of commits.  Initially, it was a single pull request with a single commit.  But then when I pushed another commit to my github repo, it piled on to the existing pull request.  I had thought that I needed to create a pull request and when I went to github, I discovered what it had done.  Do you know how to prevent github from doing this?

mooredan:
Looks like I'm not the only one to experience this problem:
https://stackoverflow.com/questions/36168839/how-to-remove-commits-from-a-pull-request
Shall I try to remedy it?

mooredan:
Ah, "in line comments" referring to the ngspice format issue, versus "in line comments" in the C source code.  ....sorry, that flew right past me.

RTimothyEdwards:
If it's not too terribly difficult, then yes, please try to do the remedy for splitting the commits into different pull requests.  I have thirteen interns this summer (trying to save my local school from the disaster of having all major summer internship programs suddenly shut down this year), so I'm a bit swamped (read: understatement).

mooredan:
Will do.

PR #22, opened 2020-06-27
Changed the end-of-line comment character from ';' to '$' when

mooredan:
formatting a SPICE netlist for NGSPICE.  This is the documented
character as described in the NGSPICE User Manual, section 2.2.4
End-of-line comments


RTimothyEdwards:
Merged.

PR #23, opened 2020-06-28
make clean deletes */Depend files as those are derived, not source

mooredan:
A little bit of housekeeping. */Depend is in .gitignore and would be removed by git clean -X


RTimothyEdwards:
I just made this fix manually and committed it.  Sorry for taking so long.

PR #25, opened 2020-06-28
During efBuildNode, each node is visited.  If a node is found to be a

mooredan:
duplicate, a couple of updates are made.  What's missing from these
updates is if the node is a "substrate" the second time around, the
EF_SUBS_NODE flag in efnode_flags needs to be set.

Then subsequent queries to this node can identify this as a substrate
node.  This is enable the removing of an earlier hack.

Quite possible not the correct solution to this problem


RTimothyEdwards:
I am not sure why this would be needed (why a substrate node would not already have the EF_SUBS_NODE flag set), but if it doesn't, then this seems like the right solution.  At least it should not do any harm that I can see.

PR #26, opened 2020-07-08
Added constraint layers to CIF Grow-min and Bridge

fabian-l-c:
Grow-min and Bridge CIF functions were modified to accept constraint layers. Modifications are:

- GROW-MIN: This function must be used to adjust tile sizes to meet a minimum width rule. This function is limited to horizontal and vertical directions (diagonal distances are not treated).
- BRIDGE: This function is used to meet minimum width (W) and minimum spacing (S) rules in the corner of tiles.

In the next example we generate implantation layers (implant_N and implant_P) from initial layers (rawN and rawP):

layer implant_N rawN
        grow-min W rawP
        bridge W S rawP
        grow S
        shrink S
calma 1 0

layer implant_P rawP
        grow-min W implant_N
        bridge W S implant_N
        grow S
        shrink S
calma 2 0

### DIFF
--- a/cif/CIFgen.c
+++ b/cif/CIFgen.c
@@ -111,6 +111,16 @@ cifPaintFunc(tile, table)
     return 0;
 }
 
+/* Data structure to pass plane and minimum width to the callback function */
+typedef struct _bridgeStruct {
+    Plane       *plane;
+    BridgeData	*bridge;
+    CellDef     *def;
+    Plane       **temps;
+} BridgeStruct;
+
+static int xOverlap, yOverlap;
+
 /*
  * ----------------------------------------------------------------------------
  *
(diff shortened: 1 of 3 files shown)


RTimothyEdwards:
Also, I don't understand the use/meaning of the additional argument to grow-min and bridge.

fabian-l-c:
Implant layers must be generated in such way that implant_N can not overlap with implant_P. For example, if we want to grow horizontally an implant_N tile to meet the minimum width specification, there are two options:

- Always grow equally quantities on both sides (half to the left and half to the right).

- Taking into account the constraint layer (in this case implant_P) the function decides in what direction to grow inplant_N. So may be the best decision is to grow only in one direction (left or right) to don't overlap the implant_N and implant_P layers.

The first option is currently implemented, the second option offers more flexibility so more cases are covered.

May be we can create another CIF operator to implement this. What do you think?

PR #31, opened 2020-07-27
Added new cif operator bridge-lim

fabian-l-c:
The new CIF operator BRIDGE-LIM is similar to the BRIDGE operator with the difference that the material created to meet the minimum spacing/width rules do not overlap the limiting layers. Syntax of the new operator is:

bridge-lim spacing width layers

where "layers" are the limiting layers.

### DIFF
--- a/cif/CIFgen.c
+++ b/cif/CIFgen.c
@@ -3769,6 +3769,831 @@ cifSrTiles(cifOp, area, cellDef, temps, func, cdArg)
 		&CIFSolidBits, func, (ClientData) cdArg);
 }
 
+/* Data structure to pass plane and minimum width to the callback function */
+typedef struct _bridgeLimStruct {
+    Plane       *plane;
+    BridgeData	*bridge;
+    CellDef     *def;
+    Plane       **temps;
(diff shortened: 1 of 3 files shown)


RTimothyEdwards:
Sorry I have been sitting on this for so long, but I had not spent any time reviewing the original "bridge" operator.  I have just come across an example where the "bridge" operator causes an overlap over a type that was being avoided by the "bloat-or" operator prior to applying "bridge", so now I can definitely see why you made this change/enhancement.  I will pull it tomorrow and review.

fabian-l-c:
OK. Let me know if there are needed modifications to put the new operator in a more convenient way. Also, I only tested with "scalefactor 1" so I am unsure about other scaling factors.

RTimothyEdwards:
@fabian-l-c :  If it's wrong, I'll find out today or tomorrow, because I will be testing it on a techfile with a non-unity scalefactor.

RTimothyEdwards:
Forgot to mention in the last comment---I have pulled and merged this request, and will now close the ticket.

Issue #32, opened 2020-07-29
MSubcircuit Extraction Method inconsistency


belohoub:
Hello!
I'm trying to get deeper into Magic's codebase, but I'm just at the beginning. As some newer features are undocumented, I'm not sure about the exact construct usages or implementation decisions - so I'm sorry for silly or redundant issue, as my understanding of the code is still strongly limited.

However, I recently experimented with "msubcircuit" extraction method in ext2spice and I found an inconsistency with the "mosfet" method.

For "mosfet", the terminal ordering is "d g s", which is correct for SPICE models, however, for "msubcircuit", the ordering is different: "s g d". It seems to me, that it might be confusing (when I use hand-written subcircuits, I prefer the same ordering for mosfet models as well as for subcircuits) - is there any reason behind?

Why are the mosfet attributes like drain/source area not available for "msubcircuit"? Those may be valuable for custom models.

RTimothyEdwards:
Generally a MOSFET drain and source will be indistinguishable, so it doesn't matter what order they are in.  Unless the device is an asymmetric MOSFET (like an extended drain device), which is a different extraction type, then magic's choice of which side of the device is the source and which is the drain is arbitrary.

That is, for a regular MOSFET, magic only distinguishes between "gate" and "other terminals", so you might get "d g s" or "s g d" for any device regardless of whether you are using the "device mosfet" or "device msubcircuit" methods.

The drain/source area are specific parameters to the M device in SPICE.  For subcircuits (X device in SPICE), you need to declare parameters that are recognized by the subcircuit, which is what the parameters in the extraction statement are for.

This is all documented;  see http://www.opencircuitdesign.com/magic/techref/maint2.html (documented but not well organized or indexed).

belohoub:
Oh, thank you for pointing me to the right document, probably I used the older version referenced here http://www.opencircuitdesign.com/magic/magic_docs.html :-(

But still, D/S are indistinguishable in general, but this is not the correct behaviour in the annotated case. In my case drain/source are annotated by using D$/S$ labels and I expect correct and consistent ordering - my point was just related to consistent ordering, especialy when Drain/source are explicitly defined. I know, that this feature was added several years ago, but stil ... it seems to me inconsistent. I guess, that "msubcircuit" (or anything) with annotated D/S should be equal to asymmetric MOSFET in the D/S uniqueness.

[...]

PR #34, opened 2020-08-09
MSubcircuit attribute extraction fix

belohoub:
This pull-request is related to issue #32 

swap Drain/Source moved into a function to encapsulate swap-related tasks; area parameters swapped according to the D/S attributes; SPICE attribute comment printf moved to the end of export function as it allows to add more device types

### DIFF
--- a/ext2spice/ext2hier.c
+++ b/ext2spice/ext2hier.c
@@ -439,6 +439,8 @@ subcktHierVisit(use, hierName, is_top)
 	return subcktVisit(use, hierName, is_top);
 }
 
+extern void swapDrainSource();
+
 /*
  * ----------------------------------------------------------------------------
  *
@@ -516,8 +518,7 @@ spcdevHierVisit(hc, dev, scale)
                 (dev->dev_terms[2].dterm_attrs &&
                 !strcmp(dev->dev_terms[2].dterm_attrs, "S")))
         {
-            drain = &dev->dev_terms[1];
-            source = &dev->dev_terms[2];
+            swapDrainSource(dev, &source, &drain);
         }
         else
             drain = &dev->dev_terms[2];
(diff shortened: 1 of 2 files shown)


RTimothyEdwards:
Mostly correct.  The search over parameters should not be searching for the "as", "ad", etc., strings;  it should just be swapping the parameter values for a1 and a2, p1 and p2.  I will pull and fix up the code.  As for the split between ext2spice.c and ext2hier.c:  Yes, there is some redundant code between the two, but I have not spent the time to try to get rid of it.  There are also a number of differences between the hierarchical and flattened output that it has to be done with a bit of care.  Mostly, though, it's just like that because I wrote ext2hier.c without knowing how easy or difficult it was going to be to get magic to write a hierarchical SPICE netlist output.  It turned out to be not as bad as I expected.  But I kept the code separate just in case I ran into a roadblock and had to start over.

RTimothyEdwards:
I realized that your code has the problem that because it is reversing the parameter definitions, they remain swapped until another swap is issued;  meanwhile, if any device is not swapped, then it gets the wrong AD, AS, PD, PS values.  So I just flagged if the swap was done, and always revert the parameters back to the original definition at the end of writing out the device.  Please check that I have not messed up the code and that it still works as you expect.  If so, I will close this pull request (Note:  Changes were updated on the opencircuitdesign.com repo master branch and will not appear in the github repo until the overnight mirror, about 14 hours from now).

belohoub:
@RTimothyEdwards Great, thank you! That make sense - I'll test your fix and let you know.

And thank you for your explenations - that make sense!

So, e.g. spcdevVisit() is completely deprecated function or it is called under some settings? It's still not clear to me (I dumped all the function calls inside magic ...) :-( 

I understand the reasons, why you left the original code for reference (or it is executed sometimes ??); and that rewriting such a complex code may lead to number of regressions, however removing the obviously deprecated code should cause no damage compared to replacing function calls (from the user perspective) - it is always possible to find it in the project history and it removes lot of confusion when studying the code ... if spcdevVisit() is really completely deprecated, then even you were confused :-)

RTimothyEdwards:
Not deprecated:  You can find it referenced in ext2spice.c lines 923 and 1090.  This is called when generating a flattened netlist (option "ext2spice hierarchy off").

belohoub:
Aha, I got it! 
It's hard for me to trace it (from some reason, I believed, that it's controlled by esDistrJunct variable leading to an undocumented switch)
 I'm sorry for the annoying comments!

Thank you!

PR #35, opened 2020-08-10
swapDrainSource reimplemented to remove side-effects

belohoub:
Thank you for fixing my proposed code - I tested it and it works!.

However, I created new pull request to solve the swapDrainSource() side-effects. The function now realy swaps the terminals (not parameters or parameter names) and causes (hopefully) no side-effects, whose must be corrected later; the "swapped" flag is thus no more required.

This code works also well.

This is related to issue #32


RTimothyEdwards:
Sorry about the longish delay, but I finally got around to pulling and merging this update.

belohoub:
Great, Thanks!

PR #41, opened 2020-10-05
Build system tweaks to make magic more relocatable

Keno:
I was trying to put together a [binarybuilder.org](https://binarybuilder.org) recipe for `magic`,
but BinaryBuilder was upset that the final result embedded lots of build system paths and
was not relocatable (i.e. being able to be moved to a different prefix after the fact).
Unfortunately, the `magic` build system appears to embed paths all over the place,
so I didn't manage to fix everything, but with the three commits in this PR, it at least
started up correctly after moving it to a different prefix.

### DIFF
--- a/scripts/configure.in
+++ b/scripts/configure.in
@@ -407,6 +407,8 @@ AC_ARG_WITH(tcllibs,    [  --with-tcllibs=DIR      Find Tcl library in DIR],
   magic_with_tcl_libraries=$withval)
 AC_ARG_WITH(tklibs,     [  --with-tklibs=DIR       Find Tk library in DIR],
   magic_with_tk_libraries=$withval)
+AC_ARG_WITH(wish,       [  --with-wish=EXE         Use wish binary at EXE],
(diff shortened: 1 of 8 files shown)


RTimothyEdwards (approved):
Sorry for the delay.  It took a while to work through all the changes, but they look good to me.  I don't particularly like the change to CAD_ROOT_DEFAULT, though.  The original intent was to look for an environment variable CAD_ROOT and copy it to a Tcl variable CAD_ROOT.  I see the need for a correction, but I would prefer to just look for Tcl variable CAD_ROOT and keep it if it exists, rather than adding an arbitrary 2nd variable CAD_ROOT_DEFAULT.  I'm going to apply your changes and then modify that.

RTimothyEdwards (commented):
I am not sure it is even possible to have a Tcl variable set at the point where you are checking for CAD_ROOT_DEFAULT.

RTimothyEdwards:
This has been pulled and merged and checked and confirmed.

Issue #45, opened 2020-11-14
Capacitance extraction for node "a" depends upon the position of label "b"


mooredan:
Cells ppcap_tc1 and ppcap_tc2 are identical in
every way except for the position of label b
on metal1.  The cell is a poly2/poly capacitor,
with node "a" being the bottom plate (poly) and
node "b" being the top plate (poly2).

Both plates are contacted to metal1.

For convenience, a screenshot of each cell is
provided:

ppcap_tc1.png
ppcap_tc2.png

![ppcap_tc1](https://user-images.githubusercontent.com/441535/99159021-bee73600-268d-11eb-9020-f9351bf76c45.png)

![ppcap_tc2](https://user-images.githubusercontent.com/441535/99159049-eb02b700-268d-11eb-92ad-d83892345624.png)

The difference in the .mag files:

```
$ diff ppcap_tc1.mag ppcap_tc2.mag
20c20
< rlabel metal1 s 19 10 19 10 2 b
---
> rlabel metal1 s 23 20 23 20 2 b

```

A TCL file "doit.tcl" is included to extract the two
cells. And to reproduce, delete the .ext files and
run the script:

```
$ rm -f ppcap{1,2}.ext
$ /usr/local/bin/magic -T SCN3ME_SUBM.30.tech -noconsole -dnull < doit.tcl > magic.log 2>&1
```

After extraction, here is the difference between the two
.ext files:

```
$ diff ppcap_tc{1,2}.ext
8c8
< port "b" 2 19 10 19 10 m1
---
> port "b" 2 23 20 23 20 m1
10c10
< node "b" 114 532.8 19 10 m1 0 0 0 0 0 0 0 0 0 0 0 0 176 60 0 0 64 40 0 0 0 0
---
> node "b" 114 532.8 23 20 m1 0 0 0 0 0 0 0 0 0 0 0 0 176 60 0 0 64 40 0 0 0 0
12,13c12,13
< cap "b" "a" 14878.1
< subcap "a" -3276.96
---
> cap "a" "b" 14878.1
> subcap "a" -1906.56
```

The significant (and electrical) difference is the subcap lines.  This is the
amount of capacitance added from the capacitance on node "a".  After translation
to a spice file, the capacitance between node "a" and the substrate
will be different.  The same design should produce the same electrical
parameters.  Label position should not affect this.

Tech file: SCN3ME_SUBM.30.tech is from
[...]

PR #46, opened 2020-11-14
Devcap2subckt

mooredan:
When active, the option transforms the device capacitors
from a "C" card to a "X" card (a subckt instance) and adds
an additional connection to the substrate node

e.g.

CN a b capmodel l=2.0u w=2.0u

becomes

XM a b subnode capmodel l=2.0u w=2.0u

The instance numbers "N" and "M" are whatever number is next
for that type of instance.

Two significant changes:

- char *substr   is made global so that it can be availabe in the
EFVisitDevs function.

- call to EFVisitNodes moved to the beginning of the routine
so that the substrate node name is found before calling
EFVisitDevs

This only affects device capacitors which have been extracted.
Future enhancement would be to designate which particular
devices it affects, and whether or not to add the substrate node.


RTimothyEdwards:
There is already an extraction device type "csubcircuit" which generates an X record for capacitor devices.

mooredan:
Thanks Tim,  Unfortunately I didn't see any documentation for csubcircuit (there's some for rsubcircuit).  I gave it a try and it worked.  What's missing is the third terminal for the substrate.  I don't propose that this be the default behavior, but would you consider a modification for adding an option to add the third terminal.  Without it, I don't see the path to adding back the capacitance between the bottom plate (poly) and the substrate.

mooredan:
Sorry, spoke too soon.  extract/ExtTech.c shows how this can be used.

I would like to do something like this:

device csubcircuit <deviceName> <types1> <types2> None $SUB <cap/area>

I used this but then $SUB shows up in the .ext file instead of what I have set SUB too:

set SUB vss

The other thing is that I'm not seeing the w= and l= parameters. :(

mooredan:
Here's the pertinent difference in the .ext file when using device capacitor vs. device csubcircuit

device devcap poly2Capacitor 101 55 102 56 16 34 "vss" 0 0 "vdd" 100 0

device csubckt poly2Capacitor 101 55 102 56 "None" "vss" 0 0 "vdd" 100 0

Note that the w= and l= values (16 34) is missing in the latter.

Once I figure out how to use this, I'll volunteer to document it for the next guy.

RTimothyEdwards:
But it is documented---see http://opencircuitdesign.com/magic ;  Click on "Technology Files" and then go to the link under "The Magic Technology File Manual".  Find the section for "extract", then scroll down to the paragraph "Device Extraction".  This shows you how to add the substrate node and how to set up parameters like length and width.

mooredan:
My oversight, I was looking here and incorrectly assumed that this was all the documentation that there was, and looked no further.

image

Well after several iterations, I correctly composed this extract section line:

```
device csubcircuit poly2Capacitor (poly2,ecap,phr,p2c)/active (poly,fp,rp,pc,pc)/active space/w l=l w=w
```
...which resulted in this line in the .ext file:

```
device csubckt poly2Capacitor 27 8 28 9 l=20 w=22 "vss" "b1" 0 0 "a1" 84 0
```
...which then resulted in the anticipated line in the spice file:

```
X0 b1 a1 vss poly2Capacitor w=6.6u l=6u
```

Issue #45 that I reported yesterday is still there though.

No need for the pull, the desired functionality is already present. Thanks for your time and assistance.

PR #47, opened 2020-11-15
maxwidth not registering DRC error when only one direction exceeds limit

mooredan:
Commit message: When comparing maxwidth, only one direction (x or y) need exceed the edgelimit parameter; versus both directions need to exceed the limit.  Changed the '&&' to a '||'

tech file maxwidth declaration in the drc section.  The DRC error is only registered when both the x and y directions exceed the limit.  By changing the logical && operator to a logical || operator, the DRC error will be registers if either direction exceeds the limit.

However, in looking at the documentation:

 ```
Maxwidth Rules

Sometimes a technology requires that a region not be wider than a certain value. The maxwidth rule can be 
used to check this.

    maxwidth layers mwidth [bends] why 

Layers, the types that compose the region, must all be in the same plane. The region must be less than mwidth wide 
in either the horizontal or vertical dimension. Bends takes one of two values, bend_illegal and bend_ok. 
For bend_illegal rules, the checker forms a bounding box around all contiguous tiles of the correct type, then 
checks this box's width.
```
...maybe this behavior is intended. If so, then may I propose an additional option to maxwidth?  That being a "both" option so that both directions are checked.

### DIFF
--- a/drc/DRCextend.c
+++ b/drc/DRCextend.c
@@ -274,7 +274,7 @@ drcCheckMaxwidth(starttile,arg,cptr)
 	    if (TTMaskHasType(oktypes, TiGetLeftType(tp))) PUSHTILE(tp);
     }
 
-    if (boundrect.r_xtop - boundrect.r_xbot > edgelimit &&
+    if (boundrect.r_xtop - boundrect.r_xbot > edgelimit ||
              boundrect.r_ytop - boundrect.r_ybot > edgelimit)
     {
 	Rect	rect;


RTimothyEdwards:
Maxwidth rules are never implemented that way.  That would imply that a skinny wire that is long enough becomes not legal---clearly that is never the case, at least not until you get down into advanced process nodes.

mooredan:
How would one implement a "must be square" rule? (there's a rect_only rule) I see that this _could_ be done by defining a width and maxwidth for a type if edgelimit in both directions were checked -- that how I got here. ....maybe there's a clever edge4way rule for this?

RTimothyEdwards:
In what context would you need a "must be square" rule?  Generally speaking, only contact cuts are required to be square, and those are handled by algorithmic generation, so there is no need for a DRC rule to check them.  That said, though, a "square_only" rule type would be a trivial matter of just copying the code for "rect_only", although I think that maybe it would work better as an option to "rect_only".  "rect_only" only uses one flag bit in the DRC rule record, as there is no other relevant information for the rule;  some other rules pass additional option information through, I think, a ClientData variable in the rule record.  e.g., set ClientData = 0 default, ClientData = 1 requires a check for a a square.

mooredan:
It is the contacts, specifically the gc2 one, but could be applied to gc and vias as well. Although the transformation of larger than minimum sized rectangles down to individual contacts is a neat feature (and I'm sure there were good reasons to do this), my personal preference is to have the individual squares.  No, I won't drawn them all by hand, but use arrays or a tcl script to "fill contacts".  I favor the WYSIWYG style when it comes to doing layout. Since I've started down the path of doing a customized tech file, I might as well make it to my personal liking.   This isn't critical, but I also like a good puzzle.

PR #48, opened 2020-11-20
CI

agorararmard:
- Dockerizing magic.
- A basic CI setup for magic that includes:
  - Documentation on how to use and add to.
  - PDK build scripts.
  - 5 test cases.
  - Test Scripts that do: GDS streaming, extractions, antenna checks, DRC checks.

Please, review when you have the time and make sure you're comfortable with the way test cases are handled and that you can add test cases or test scripts to this in the future (documentation exists inside .travisCI).

P.S: I'm not the best magic user. So, I'm open to all changes that you require.


RTimothyEdwards (approved):
Since this apparently doesn't interfere with any of the code base, then I'll go ahead and approve it, although I won't be able to actually take a look at it closely until after the tape-out, I expect.

RTimothyEdwards:
This has been merged into the master branch on opencircuitdesign.com, thanks!

PR #49, opened 2020-11-22
Extends maxwidth DRC rule check to optionally check maxwidth in

mooredan:
both directions. Adds one more valid option to the optional "bends"
argument.  If "both" is specified, then the maximum width will
be checked in both directions (the default is to check if only
one direction exceeds the maxwidth distance).

Existing functionality is preserved and does not force a change
to any existing design or tech file.

If combined with a width rule, then an "square_only" rule can
be had, e.g.:

 width (genericpoly2contact,gc2)/contact 2 \
 "Exact contact size is 2x2 (Mosis #13.1) - AMI C5X CNTW (0.5 um)"

 maxwidth (genericpoly2contact,gc2)/contact 2 both \
 "Exact contact size is 2x2 (Mosis #13.1) - AMI C5X CNTW (0.5 um)"


mooredan:
Had conflicts -- mergeable again.

RTimothyEdwards:
Pulled and merged.  Sorry about the long delay.

PR #53, opened 2020-12-06
Extract: After subtracting cap from running total, if the total

mooredan:
is between -1 and 1 then set the total to zero.  The units of the C on the node line (fourth field) is in attofarads.

This "residual" comes about due to arithmetic resolution.  This small "zero" number then can end up in the .ext file

After extraction, noticed this line in the .ext file:

`node "b" 114 -5.68434e-14 180 210 ecap 0 0 0 0 0 0 0 0 0 0 158400 1800 0 0 8100 360 0 0 0 0`

Note the minuscule negative capacitance number.  After applying the patch, this line becomes:

`node "b" 114 0 180 210 ecap 0 0 0 0 0 0 0 0 0 0 158400 1800 0 0 8100 360 0 0 0 0`


mooredan:
In the pull request, the change amounts to testing if the sum (nreg_cap, a double) is -1 > sum < 1 (effectively zero since the units of  are attofarads).  Maybe tightening this up to -0.001 < sum < 0.001 (or smaller) would be more conservative?  Many discussions on how to compare floats to zero, including this one: https://stackoverflow.com/questions/19837576/comparing-floating-point-number-to-zero

RTimothyEdwards:
Pulled and merged.

PR #54, opened 2020-12-07
Ignore derived files when compiling on OSX

mooredan:
Add a file and a couple of directories to the .gitignore. These are derived when compiling on OSX


RTimothyEdwards (approved):
Don't know much about OS-X but I'll take your word for it.

RTimothyEdwards:
Pulled and merged.

PR #56, opened 2020-12-23
Update config.guess/config.log

antonblanchard:
The current version doesn't understand ppc64le Linux.


RTimothyEdwards (approved):
Too big to  review but I agree that an update of config.guess and config.sub is due.  Or overdue.

RTimothyEdwards:
Pulled and merged.

PR #57, opened 2020-12-23
Fix a few compiler warnings

antonblanchard:
We were missing string.h in a few places, and were also creating
a string format of %llld.


RTimothyEdwards (approved):
Looks good.

RTimothyEdwards:
Pulled and merged, thanks!

PR #59, opened 2020-12-27
CIFgen: Fix processing of MASKHINTS_* properties

smunaut:
- Don't print error if there is no more data to process
 - Fix error message: `propname` already has the
   MASKHINTS_ prefix

Signed-off-by: Sylvain Munaut <tnt@246tNt.com>


RTimothyEdwards (approved):
All of that makes perfect sense;  thanks for the fixes!

RTimothyEdwards:
Pulled and merged (to master branch at opencircuitdesign.com).

PR #61, opened 2021-01-03
Set GrClosePtr back to NULL before calling MainExit()

mooredan:
to squash the seg fault. It appears that calling GrTkClose
before graphics are fully initialized causes this.


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #65, opened 2021-01-14
Propagate errors when writing a GDS

ax3ghazy:
Would you like me to make this behaviour optional? (e.g., `gds allowerrors true` or so)


RTimothyEdwards (commented):
It can't be written this way, but I just implemented it and pushed the update.  I ran some simple tests to confirm that it works.

Note that your code change would prevent magic from reading in any database with a missing cell, as it would halt the read-in after the first cell not found, which is not the intended behavior of reading a .mag file with undefined references.  It should only affect GDS writes.  I decided to retain the behavior of halting the search on extraction, DRC checks from the top, and LEF and CIF writes as well.  Reading a database will not halt on failure to find a cell.

RTimothyEdwards:
@ax3ghazy :  Please pull my implementation and check that it works as expected.

ax3ghazy:
The latest version produces the desired behaviour; closing this.

PR #68, opened 2021-02-18
Restore terminal if exit is called inside a TCL script at startup

antonblanchard:
If exit is called in a TCL script that is executed at startup, the libc
exit() function is called directly and we don't get a chance to reset
the terminal. We return to the shell with echo off, and have to run
"reset". A simple example:

echo exit > test.tcl
magic -noconsole -dnull  test.tcl

There are a few ways we could solve this. We could register an exit
handler using atexit(). Here I use Tcl_SetExitProc() to register a
callback with the TCL interpreter.

### DIFF
--- a/utils/main.c
+++ b/utils/main.c
@@ -770,6 +770,12 @@ mainInitAfterArgs()
     return 0;
 }
 
+void tcl_exit_hook(ClientData clientData)
+{
+    TxResetTerminal();
+    exit(0);
+}
+
 /*
  * ----------------------------------------------------------------------------
  * mainInitFinal:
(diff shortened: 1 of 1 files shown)


RTimothyEdwards:
I use tcsh as my shell so I have not seen this behavior before.  Since it is quite annoying on bash, I will go ahead and apply this update (and make sure it doesn't have any unwanted side effects on tcsh).

RTimothyEdwards:
Pulled and merged (to opencircuitdesign.com;  will mirror to github in about 6 hours).

antonblanchard:
Thanks @RTimothyEdwards!

PR #75, opened 2021-05-03
Fix build error with clang

Keno:
At least in some configurations of clang, `-Wreturn-type` is set to
error by default, causing the magic build to fail with:

```
ExtCell.c:115:2: error: non-void function 'ExtCell' should return a value [-Wreturn-type]
```


RTimothyEdwards:
Sorry, somehow this escaped my attention.  Fixing the code now.

RTimothyEdwards:
Fixed on opencircuitdesign.com master branch, will update on github overnight.

PR #98, opened 2021-11-16
Fixed two bugs on the `def write` command

mbalestrini:
- def write was setting x,y location of `COMPONENTS` using the boundaries of the cell, without taking into account it's internal origin point
- The warning message about nets not having the default width was showing the wrong dimension (the bigger one instead of the smaller)


RTimothyEdwards (approved):
Looks good to me.  Thanks for looking at the "def write" code, which doesn't get much attention.

RTimothyEdwards:
I'll get this pulled and merged tomorrow---there are changes I made to the magic code today that are pending.

RTimothyEdwards:
Pulled and merged.

PR #113, opened 2021-12-30
Correct & has lower precedence than parentheses warnings.

brad0:
CmdRS.c:1269:22: warning: & has lower precedence than ==; == will be evaluated first [-Wparentheses]
DRCtech.c:2573:16: warning: & has lower precedence than !=; != will be evaluated first [-Wparentheses]


RTimothyEdwards (approved):
Yes, good catches.

RTimothyEdwards:
Pulled and merged.

PR #115, opened 2021-12-31
Add AppImage Building Capabilities

donn:
This adds a feature that builds AppImages (https://appimage.org/) that can produce a monolithic magic binary.

Demo: `curl -L https://github.com/donn/magic/releases/download/donns-test/Magic-donns-test-x86_64.AppImage > magic && chmod +x magic && ./magic -dnull -noconsole`

This should theoretically work on any Linux distro with Glibc newer than 2.3, i.e., any GNU/Linux released in the last 10 years.

I also added a CI that automatically builds and creates a release for said binary.


RTimothyEdwards (commented):
Does it have to have a hard-coded Tcl/Tk version?  Why does it download Tcl/Tk from sourceforge when it doesn't do that for any other of the packages it uses?

donn:
Tcl/Tk is the only one library that is not in *every* Linux installation dynamically links against. That's why they've been built from source. Note that I don't go for GL or Cairo in this build.

If you inspect the binaries with `ldd,` the rest are ABI-stable dylibs that are common across the vast majority of GNU-based Linux distros.

RTimothyEdwards:
Not including OpenGL or Cairo kind of makes magic crippleware, doesn't it?

donn:
Not for CLI use. That said, if you'd rather have me figure that GL/Cairo first, that's also fine.

RTimothyEdwards (approved):
As it has its useful purposes, I'll approve it.  I can see it as a long-term pain, particularly with hard-coded references to specific Tcl/Tk tarballs and the like.

donn:
Wait, Tim, let's not merge this for now if possible. I still need to clean it up but I've been so busy the last while.

RTimothyEdwards:
But it won't harm anything if merged, right?  It would have to have something enabled in github workflows, right?

donn:
No, as long as you have any actions enabled, which you do, this action will automatically be registered and will run. Not ideal.

RTimothyEdwards:
It is only some added files.  I'll hold off merging until you okay it, then.

donn:
I added support for Cairo 1.8+.

I'm not going to build and include it with magic like Tcl/Tk as it's a considerably more complex build.

RTimothyEdwards:
Okay, but is it ready to pull and merge?

RTimothyEdwards:
@donn :  I could not merge this PR cleanly.

PR #116, opened 2022-01-01
fix tech file load error with qflow

susinxy:
issue:
qflow passes tech file name without the '.tech' suffix. however magic tech load function treats any filename with suffix which in not '.tech' as a specified filename should be honored, leading to a problem that fails to load osu tech file

fix:
try loading a honored file with suffix '.tech' if the honored file is not there

### DIFF
--- a/utils/tech.c
+++ b/utils/tech.c
@@ -476,8 +476,13 @@ TechLoad(filename, initmask)
 
 	/* If a non-standard extension was used, then honor it */
 	if ((dptr != NULL) && (*dptr != '\0'))
+    {
 	    tf = PaOpen(filename, "r", (char *)NULL, ".", SysLibPath, &realname);
-	else
+        /* If the honored file in not there, try the filename with the suffix */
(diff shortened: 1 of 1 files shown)


RTimothyEdwards:
This seems like the wrong solution---although clearly something is going on.  If qflow passed the techfile name without the ".tech" suffix, then there is no unique extension to honor at all.  So the modified code shouldn't even be running.  I'll look into it.

RTimothyEdwards:
Okay, I get it.  It's the use of the superfluous "." in the technology name, e.g., `SCN4M_SUBM.20.tech`, which then still looks like it has an extension when the `.tech` is stripped off the end.

RTimothyEdwards (approved):
Okay, this does work for the unwisely named technology files like `SCN4M_SUBM.20.tech`.  Thanks for tracking down the issue.

RTimothyEdwards:
This should be fixed now, so I'm going to go ahead and close the issue.  Feel free to re-open it if there are still any problems.

PR #117, opened 2022-01-07
Fix typo in MakeLegalLEFSyntax

antonblanchard:
In MakeLegalLEFSyntax we step through the badLEFchars string
but instead of looking for the NULL terminator, we instead
look for a NULL pointer.


RTimothyEdwards (approved):
Good catch.

RTimothyEdwards:
Pulled and merged, thanks.

PR #118, opened 2022-01-07
Add missing includes

antonblanchard:
A number of places are using isspace(), tolower(), toupper() and strcmp()
without including the relevant header.


RTimothyEdwards (approved):
Thanks, looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #119, opened 2022-01-07
Add missing types in function prototypes

antonblanchard:
It would be nice to convert the codebase from K&R to ANSI function
prototypes, but for now just clean up the warnings.


RTimothyEdwards (approved):
All well and good, although it bothers me that gcc will compile source code that is missing type declarations in a subroutine.

RTimothyEdwards:
Pulled and merged.

PR #120, opened 2022-01-11
Add more information to illegal overlap warnings

antonblanchard:
Print the names of the tile types that are illegally overlapping. This
gives us a better idea of what is wrong, eg:

feedback add "Illegal overlap between obsm2 and m2 (types do not connect)" medium


RTimothyEdwards:
Pulled and merged.

PR #132, opened 2022-02-16
fix errors in cmdDumpParseArgs ( getcell command error)

Park-Hyung-Joo:
there are some errors in getcell command
error case1: getcell cellname child 0 0 parent ll -> read 0 as label
why: else if( ac ==3 ){ coordiate case } -> since ac = argc -2 at first -> argc > 3 -> pass else if() and go to label case
sol: (ac >= 3) and check if following two strings are number -> satisdfy coordinate case
error case2: getcell cellname v 0 0 -> not transform ( transform default case executed )
why: '0', '90', '180', '270' are in kwdNames -> else if ( Lookup(av[1], kwdNames)>=0 ) { go to default } executed
sol: these command never invoked before function checks transform option '0'(no transform) or '90'(rotate)...
so just check if av[1] are one of them, then just pass "else if ( )"

### DIFF
--- a/commands/CmdCD.c
+++ b/commands/CmdCD.c
@@ -4573,6 +4573,7 @@ cmdDumpParseArgs(cmdName, w, cmd, dummy, scx)
      * points weren't provided.  (Lower-left of the box tool is interpreted
      * in root coordinates).
      */
+	// getcell cellname child 0 0 parent ll v 0 0
     av = &cmd->tx_argv[2];
     ac = cmd->tx_argc - 2;
     hasChild = hasRoot = hasTrans = FALSE;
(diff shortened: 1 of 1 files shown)


RTimothyEdwards (approved):
Looks okay to me.  The last one with the multiple strcmp(...) calls looks clumsy, but as long as it works, I'm okay with it.

Park-Hyung-Joo:
Thxs. strcmp... I think so too, but it would be better than Lookup(av[1], kwdNames) != 2 && Lookup(av[1], kwdNames) !=3 ...

RTimothyEdwards:
Yes, but you could do `result = Lookup(av[1], kwdNames);` before the conditional and then use `if ..  result >=2 && result <= 4`.  But it's not really all that important.

Park-Hyung-Joo:
I see. Thank you for your suggestion

PR #135, opened 2022-02-23
Fix a few issues with missing or incorrect prototypes

antonblanchard:
There are a few places we either don't have a variable or
function prototype and need one, or we have one and they
don't match.


RTimothyEdwards (approved):
Doing a push anyway, so I'll merge this in with the rest.

RTimothyEdwards:
Pulled and merged.

PR #143, opened 2022-03-17
scripts/configure: enable building without x

proppy:
- use `no_x` instead of `X_DISPLAY_MISSING` (which is only defined as a compiler flag)
- make `X11` related error message non fatal
- only add `X11THREAD_SRCS` if `X11` is enabled.

### DIFF
--- a/scripts/configure
+++ b/scripts/configure
@@ -6058,10 +6058,10 @@ fi
 fi
 
 
-if test "x${X_DISPLAY_MISSING}" = "x"; then
+if test "x$no_x" = "x"; then
   usingX11=1
 else
-  as_fn_error $? "Unable to find X11---compiling without graphics" "$LINENO" 5
+  echo Unable to find X11---compiling without graphics.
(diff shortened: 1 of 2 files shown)


proppy:
@RTimothyEdwards should we re-generate the configure file rather than duplicating the edits?

RTimothyEdwards:
Sorry, I've been busy and hadn't looked at this yet.  I had assumed that you did regenerate the configure file;  but it doesn't matter to me.  I'll just pull and merge and then re-run autoconf.

RTimothyEdwards (approved):
I have checked both changes and yes, they are needed.

RTimothyEdwards:
Pulled and merged in the opencircuitdesign.com repository;  will be mirrored to github in a few hours.

proppy:
@RTimothyEdwards re-running `autoconf` generated a bunch of unrelated changes (probably due to the version of autoconf being more recent) so I was relunctant to bring all those changes in at the same time (as they could complicate the review).

PR #155, opened 2022-04-09
Use the python3 that configure found

ryandesign:
When running preproc.py, use the python3 that the configure script found instead of assuming (via preproc.py's `#!` line) that it is called "python3" and is located in PATH. This allows the user to specify a different python3 by running e.g.:

```
./configure ac_cv_path_PYTHON3=/path/to/python3
```

### DIFF
--- a/scripts/configure
+++ b/scripts/configure
@@ -5367,7 +5367,7 @@ if test "x${PYTHON3}" == "x"; then
    fi
 
 else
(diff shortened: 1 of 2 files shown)


RTimothyEdwards (approved):
Works for me.

RTimothyEdwards:
Pulled and merged.

PR #158, opened 2022-04-11
Fix inadequate quoting in MSED variable

ryandesign:
Fixes:

```
sed: 1: "s/\/\\/
": unescaped newline inside substitute pattern
```


RTimothyEdwards:
Pulled and merged.

PR #161, opened 2022-04-29
aarch64 fix + linux/mac build CI

donn:
+ Add "smoke test" build CI to Github Actions
+ Added a configure_mac script (requires brew)
~ caddr_t -> void* (was never a part of any UNIX standard)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,22 +4,15 @@ name: CI
 
 # Controls when the workflow will run
 on:
-  # Triggers the workflow on push or pull request events but only for the master branch
   push:
-    branches: [ master ]
   pull_request:
-    branches: [ master ]
-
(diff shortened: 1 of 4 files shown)


donn:
So this works on aarch64 Monterey but not amd64 Big Sur. I give up. Let's just merge this as-is.

saicharan0112:
@donn I believe smoke test is already present in the existing CI. It checks for basic build and functionality. Is this what you added in this pull request?

donn:
Not for macOS (which I've disabled for now, admittedly). And these are some basic adjustments to get Magic to compile better on macOS.

saicharan0112:
oh, ok. Got it!

RTimothyEdwards (approved):
Looks good.

RTimothyEdwards:
Pulled and merged.

PR #169, opened 2022-06-13
Fix bugs in port first, next commands.

mguthaus:
Previous commits changed the port IDs to an integer rather than a
bitfield. However, the first and next commands were utilizing that
a -1 became a large positive integer when masked. This resulted in
the min port operations failing. Added a default comparison with -1
to fix the problem.


RTimothyEdwards (approved):
Yes, looks good, thanks.

PR #180, opened 2022-08-24
Antenna checker should ignore vias in partial mode

antonblanchard:


RTimothyEdwards (commented):
I'm going to reject this change, not because it's wrong, but because I want to move the check for PARTIAL mode into the antenna area accumulation function, since that prevents it from running all the calculations on the contact area in the first place.  I'll post a notification of the update in a minute.

RTimothyEdwards:
I have implemented the correction per my comment above and pushed to the repo.  Thanks for the fix!

RTimothyEdwards:
This fix got lost somehow.  Trying again.

RTimothyEdwards:
I reviewed the antenna checking code yesterday and determined that there were several errors around checks of nets connected to diode diffusion.  Once was that it did not make an exception for partial mode, like the code for this PR, but also it was double-counting the area of contacts in the diffusion area calculation, and it was not treating the diffusion area ratio correction factor as a dimensioned unit.  All of these errors have now been checked and corrected.

PR #181, opened 2022-09-03
Make use of termios on OpenBSD

brad0:


RTimothyEdwards (approved):
This all looks good to me.

RTimothyEdwards:
Pulled and merged.

Issue #69, opened 2021-02-20
Implicit declaration of function is invalid in C99


jay20162016:
When building magic on mac, I get 
```
--- errors and warnings logged in file make.log
--- making header file database/database.h
--- making dependencies
--- making modules
--- compiling bplane/bpEnum.o
In file included from bpEnum.c:39:
In file included from bpEnum.c:39:
bpEnum.c:76:5: error: implicit declaration of function 'GeoCanonicalRect' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
--- compiling cmwind/CMWcmmnds.o
In file included from CMWcmmnds.c:32:
In file included from CMWcmmnds.c:32:
--- compiling commands/CmdSubrs.o
In file included from CmdSubrs.c:33:
In file included from CmdSubrs.c:33:
CmdSubrs.c:409:14: warning: using the result of an assignment as a condition without parentheses [-Wparentheses]
CmdSubrs.c:409:14: note: place parentheses around the assignment to silence this warning
CmdSubrs.c:409:14: note: use '==' to turn this assignment into an equality comparison
CmdSubrs.c:433:9: error: implicit declaration of function 'DBTechNameTypes' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
CmdSubrs.c:433:9: note: did you mean 'DBTechNameType'?
CmdSubrs.c:755:8: warning: using the result of an assignment as a condition without parentheses [-Wparentheses]
CmdSubrs.c:755:8: note: place parentheses around the assignment to silence this warning
CmdSubrs.c:755:8: note: use '==' to turn this assignment into an equality comparison
--- compiling database/DBbound.o
In file included from DBbound.c:26:
In file included from DBbound.c:26:
DBbound.c:61:9: error: implicit declaration of function 'DBSrCellPlaneArea' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
--- compiling dbwind/DBWcommands.o
In file included from DBWcommands.c:29:
In file included from DBWcommands.c:29:
--- compiling dbwind/DBWdisplay.o
In file included from DBWdisplay.c:34:
In file included from DBWdisplay.c:34:
DBWdisplay.c:652:5: error: implicit declaration of function 'GrBox' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
DBWdisplay.c:815:2: error: implicit declaration of function 'GeoTransPointDelta' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
[...]

PR #191, opened 2022-10-10
Compatibility with C99 standard

just22:
Dear Tim,

This diff makes the code (mostly) C99-compatible, enabling to compile it without the -Wno-error=implicit-function-declaration flag. This way, Magic becomes usable on arm64 architectures, specifically on Apple computers with M1/M2 SoC.

The diff is gigantic (more than 150 files modified), but rather simple; I marked each modification with the `/* C99 compat */` comment, to make it easier to review even after committing.

This addresses (and should resolve) #69; I lightly tested Magic on a M1 Macbook and can confirm that all the segmentation faults I was experiencing before have gone away; in particular, I'm now able to compile the Skywater-130 PDK using open_pdks scripts.

Please let me know in case of any doubt.

Cheers

--
Alessandro

### DIFF
--- a/calma/CalmaRdcl.c
+++ b/calma/CalmaRdcl.c
@@ -46,6 +46,9 @@ static char rcsid[] __attribute__ ((unused)) = "$Header: /usr/cvsroot/magic-8.0/
 #include "calma/calmaInt.h"
 #include "calma/calma.h"
 
+/* C99 compat */
+#include "drc/drc.h"
+
 int calmaNonManhattan;
 int CalmaFlattenLimit = 10;
 int NameConvertErrors = 0;
--- a/calma/CalmaRdio.c
+++ b/calma/CalmaRdio.c
@@ -46,6 +46,9 @@ static char rcsid[] __attribute__ ((unused)) = "$Header: /usr/cvsroot/magic-8.0/
 #include "textio/textio.h"
 #include "calma/calmaInt.h"
 
+/* C99 compat */
+#include "calma/calma.h"
+
 /* Forward declarations */
 bool calmaReadR8();
 bool calmaSkipBytes();
--- a/calma/CalmaRdpt.c
+++ b/calma/CalmaRdpt.c
@@ -48,6 +48,9 @@ static char rcsid[] __attribute__ ((unused)) = "$Header: /usr/cvsroot/magic-8.0/
 #include "calma/calmaInt.h"
 #include "calma/calma.h"
 
+/* C99 compat */
+#include "drc/drc.h"
+
 extern int calmaNonManhattan;
 
 extern int CalmaPolygonCount;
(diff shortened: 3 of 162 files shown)


just22:
Hello Tim,

Any comments on this PR?

Cheers

--
Alessandro

RTimothyEdwards:
Yes, I'm way behind on dealing with pull requests.  Give me a few more days.

donn:
@just22 I'd advise adding `-Werror=implicit-function-declaration` to the compiler flags just so future implicit declarations error out on GCC as well.

just22:
@donn Seconded. But I let Tim comment on that and decide where to enforce the flag.

Cheers

--
Alessandro

RTimothyEdwards (changes requested):
Okay, finally getting around to looking at this.
Very valuable code fixes. . . something that I have balked at for years because it requires changes to 150 files!
Anyway, it all looks good except for the following four items:
1) grTkCommon.c:  There should not be references to OpenGL here, so no need to include anything from OpenGL.  Where is that dependency coming from?  At very least, if it is required, it should be wrapped in `#ifdef OGL . . . #endif` (try compiling without OpenGL support to see what breaks).
2) grTkCommon.c:  I think the calls to `panic()` were inadvertently copied from some code example using them.  I'd rather see the `panic()` calls replaced than to add an include for `mach/mach.h`.  Probably `panic()` should just be replaced with `TxError()` to print the error, followed by `MainExit()` to exit somewhat more gracefully.  At very least, it should be `Tcl_Panic()` and not `panic()`.
3) textioInt.h:  Any routine or variable with `Tcl_` in front should be wrapped in `#ifdef MAGIC_WRAPPER . . . #endif`.  Try compiling without Tcl/Tk support and see what breaks.
4) graphicsInt.h:  Any variable containing "cairo" should be wrapped in `#ifdef CAIRO . . . #endif`.  Try compiling without cairo graphics support to see what breaks.

RTimothyEdwards:
@donn, @just22 :  I have no issue with having `-Werror=implicit-function-declaration` in the build flags, once it no longer stops magic from compiling.

just22:
Dear Tim,

Thanks a lot for your comments and sorry for the late reply, but applying your suggestions has been more complex than expected, since disabling TCL support revealed a bunch of further implicit functions declarations to fix.

Now I have reviewed the compile phase with and without TCL support, with and without CAIRO support, with and without OGL support and with and without X support: all looks good.

Please review all new commits and let me know.

Note: I didn't add `-Werror=implicit-function-declaration` to the build flags: I let you do that, once you're convinced that the code compiles in each and every case.

Cheers

--
Alessandro

RTimothyEdwards (approved):
This looks good now;  I'm going to try it out, and then push if it all looks good.

RTimothyEdwards:
@just22 :  I merged this code, then found a few errors on my end and fixed them, and then worked on a handful of warnings that were exposed (previously buried among the slew of implicit function declaration warnings).  I then went ahead and added in the "-Werror=implicit-function-declaration" to the CFLAGS in the configuration script.  Please test the resulting code at your convenience.  And thanks for the hard work!  (And any time you feel like tackling "implicit-int". . . )

Issue #199, opened 2022-11-12
C99 implicit declaration of function warning


brad0:
After the recent commits for C99 compat, doing a build on OpenBSD I see one implicit declaration of function warning..

magic 8.3.340

```
cc -O2 -pipe -Wno-parentheses -Wimplicit-int -fPIC  -I/usr/local/include/tk8.5 -I/usr/local/include/tcl8.5 -I. -I.. -I/usr/local/include -I/usr/X11R6/include -DCAD_DIR=\"/usr/local/lib\" -DBIN_DIR=\"/usr/local/bin\" -DTCL_DIR=\"/usr/local/lib/magic/tcl\"  -DUSE_TCL_STUBS -DUSE_TK_STUBS -DPACKAGE_NAME=\"\" -DPACK
AGE_TARNAME=\"\" -DPACKAGE_VERSION=\"\" -DPACKAGE_STRING=\"\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DSTDC_HEADERS=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_MEMORY_H=1 -DHAVE_STRINGS_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_UNISTD_H=1 -DSIZEOF_VOID_P=8 -D
SIZEOF_UNSIGNED_INT=4 -DSIZEOF_UNSIGNED_LONG=8 -DSIZEOF_UNSIGNED_LONG_LONG=8 -DSTDC_HEADERS=1 -DHAVE_SETENV=1 -DHAVE_PUTENV=1 -DHAVE_SYS_MMAN_H=1 -DHAVE_DIRENT_H=1 -DHAVE_LIMITS_H=1 -DHAVE_PATHS_H=1 -DHAVE_VA_COPY=1 -DHAVE___VA_COPY=1 -DHAVE_ZLIB=1 -DFILE_LOCKS=1 -DCALMA_MODULE=1 -DCIF_MODULE=1 -DPLOT_MODULE=1 -
[...]

PR #208, opened 2022-12-28
Add PlowShowTime prototype when NO_RUSAGE isn't defined

just22:
Hello Tim,

This commit adds a missing function prototype in plow module; while there, I changed the function type from (implicit) int to void, since it doesn't return any value.

It should address #199 (only compiled on OpenBSD, no real test done).

--
Alessandro

### DIFF
--- a/plow/PlowMain.c
+++ b/plow/PlowMain.c
@@ -2249,6 +2249,7 @@ plowYankCreate()
 }
 
 #ifndef	NO_RUSAGE
+void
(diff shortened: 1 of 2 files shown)


RTimothyEdwards (approved):
Looks good to me.  Sorry for not getting around to merging it for a couple of weeks.

RTimothyEdwards:
Pulled and merged (but not yet mirrored to github).

brad0:
@just22 Thank you.

PR #209, opened 2023-01-07
Use offsetof()

antonblanchard:
Magic has an open coded version of offsetof() which clang complains about, because it takes the offset of a null pointer. Use offsetof() instead.

### DIFF
--- a/bplane/bpMain.c
+++ b/bplane/bpMain.c
@@ -37,6 +37,7 @@
  */
 
 #include <stdio.h>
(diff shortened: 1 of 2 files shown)


RTimothyEdwards (approved):
The bplane code was merged from the open source micromagic distribution, and I never paid any attention to this part of the code because it works fine on gcc.

RTimothyEdwards:
Pulled and merged (but not yet mirrored to github).

PR #217, opened 2023-02-02
Remove cast of pointer to struct and back again.

antonblanchard:


RTimothyEdwards:
Pulled and merged (will get mirrored to github overnight).

Issue #195, opened 2022-11-01
Macos: typedef redefinition with different types ('struct reg' vs 'struct _XRegion *')


cheyao:
I've installed everything according to https://github.com/RTimothyEdwards/magic/blob/master/INSTALL_MacOS.md

configure mac:
```
-----------------------------------------------------------
Configuration Summary (principle requirements):

X11:          yes
Python3:      yes
OpenGL:       no

  Cairo graphics are available so OpenGL is purely optional.
Cairo:        yes
Tcl/Tk:       yes
-----------------------------------------------------------

Use 'make' to compile and 'make install' to install.

-----------------------------------------------------------
```
and then running make -j$(sysctl -n hw.ncpu) gives this error:

```
[...]

PR #222, opened 2023-02-24
rename Region to NetRegion

xobs:
On Macos, Tk includes X11, which has this define:

    #if !defined(MAC_OSX_TK)
    #   include <X11/X.h>
    #endif
    #ifdef MAC_OSX_TK
    #   include <X11/X.h>
    #   define Cursor XCursor
    #   define Region XRegion
    #endif

This causes the keywords `Cursor` and `Region` to be reserved, which prevents magic from building (see #195)

Rename `Region` to `NetRegion` as suggested by @RTimothyEdwards in https://github.com/RTimothyEdwards/magic/issues/195#issuecomment-1298870983


RTimothyEdwards (changes requested):
I would like to go ahead and merge this.  However, can you please change "NetRegion" to "ExtRegion", in keeping with magic's usual nomenclature.  There is a subdirectory "net" with variables prefixed with "Net", so the proper thing to do here is prefix the variable with the same name as the directory, which is "Ext".

xobs:
Alright, I performed the rename. The branch isn't correctly named anymore, so let me know if you'd like me to open a new PR.

RTimothyEdwards (approved):
Looks good to me.  There are a couple of places where the blanket replacement caught some comments that were unrelated to the structure name.  I just fixed those by hand.

PR #256, opened 2023-07-11
add wasm target

proppy:
- add new `asmjs-unknown-emscripten` target
- add `defined(EMSCRIPTEN)` guards
- use `va_list` for variadic function call forwarding
- fix fatal function arguments inconsistency
- add `simple_build_wasm` CI job and store main `magic` wasm bundle as an artefact

Note: currently building with `--without-cairo --without-opengl --without-x --disable-readline` and bundle is already quite big (10M), any idea on how to trim things further?

/cc @urish since this might be relevant for https://github.com/wokwi/siliwiz

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,7 +49,7 @@ jobs:
       - name: Build
         run: |
           source ./emsdk/emsdk_env.sh
-          emconfigure ./configure --without-cairo --without-opengl --without-x --disable-readline
+          emconfigure ./configure --without-cairo --without-opengl --without-x --disable-readline --target=asmjs-unknown-emscripten
           emmake make
       - name: archive wasm bundle
         uses: actions/upload-artifact@v3
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,6 +35,28 @@ jobs:
           ./configure
           make database/database.h
           make -j$(nproc)
+  simple_build_wasm:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Get Dependencies
+        run: |
+          sudo apt-get install -y csh
+          git clone https://github.com/emscripten-core/emsdk.git
+          cd emsdk
+          ./emsdk install latest
+          ./emsdk activate latest
+      - name: Build
+        run: |
+          source ./emsdk/emsdk_env.sh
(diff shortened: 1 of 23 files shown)


RTimothyEdwards (changes requested):
Looks good except that it looks like you missed having TxError() call TxErrorV().  Unless I'm missing something, it is just running va_start() and va_end() and then returning.

proppy:
> Looks good except that it looks like you missed having TxError() call TxErrorV(). Unless I'm missing something, it is just running va_start() and va_end() and then returning.

Good catch! Fixed :)

RTimothyEdwards (approved):
Looks good to me.

proppy:
@RTimothyEdwards any ideas on how to trim the bundle further? (any other thing we could disable?)

RTimothyEdwards:
I would not consider 10M very big at all for something as expansive as Magic.  If you are building without graphics, then a lot of the files in the target install `lib/magic/sys/` directory are not needed, including the vector fonts and the default tech files.  I don't think that saves you very much space, though.

proppy:
@RTimothyEdwards would building `libmagic` rather than `magic`'s `main` save anything?

RTimothyEdwards:
Pulled and merged (in opencircuitdesign.com).

proppy:
@RTimothyEdwards thanks as always (and we missed you this year at FSiC!)

RTimothyEdwards:
@proppy :  No, I don't think `libmagic` will do anything meaningful.
What is the main contributor to the size of your bundle?  The shared object library?

(Yeah, sorry I couldn't make FSiC.  I missed seeing all my European (and Asian) friends.  I will try to make it a regular yearly habit, and this year should be just a one-off exception.)

proppy:
@RTimothyEdwards PTAL, added https://github.com/RTimothyEdwards/magic/pull/256/commits/26d2ceb77fd7622e06691e2f76baddce9110d0ca which fix a warning/error I got at runtime when testing the bundle (let me know if you'd prefer a separate PR since you already pulled on opencircuitdesign.com).

RTimothyEdwards:
@proppy :  Yes, actually I would like that as a separate PR, although I can also just manually fix it, as that's probably less involved than making a new PR.

proppy:
Closing since this is merged, will send another PR for the fixups.

PR #258, opened 2023-07-16
Build Updates

donn:
This is a series of updates to make building magic far less of a headache:

* Drop csh/tcsh dependency and detection from ./configure
* Rewrite makedbh in Python (confirmed 1:1 output)
* Rewrite printmans in POSIX sh
* Stop deleting Depend before every compile (which causes some files to recompile and thus increases recompile times significantly)
* Add Depend to CLEANS in scripts/defs.mak.in
* Turn POSIX suffix rule in magic/rules.mak to a pattern rule with proper prerequisites

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,10 +17,10 @@ jobs:
     steps:
       - name: Pulling the docker image
         run: docker pull vezzal/vezzal:v1
-        
+
       - name: Start the container with the docker image
         run: docker run -id --name test_magic vezzal/vezzal:v1 bash | exit
-        
+
       - name: Run the testing on the container and send the mail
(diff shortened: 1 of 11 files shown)


RTimothyEdwards (approved):
All of this looks great.  Sorry for taking so long to get around to it.

RTimothyEdwards:
@donn :   There is one remaining issue which is that if I do "make clean" more than once, then the 2nd time it starts trying to compile everything again.  Apparently there's some dependency on "make clean" that makes it want to generate everything first?

RTimothyEdwards:
@donn :  I'd appreciate it if you could look into why it is no longer possible to do "make clean" more than once without running into a slew of errors due to database/database.h having been removed.  It seems to have to do with moving the "Depend" into the "make clean" recipe's dependency list.

PR #266, opened 2023-08-29
Allow Python 3 to not be in PATH, Fix `clean` re-building dependencies

### LINKED ISSUE #265: env: python3: No such file or directory
magic 8.3.425 fails to build on older systems that don't have a `python3` executable, like macOS 10.14 and earlier:

```
--- making header file database/database.h
./scripts/makedbh database/database.h.in database/database.h
env: python3: No such file or directory
make: *** [database/database.h] Error 127
```

This problem is new in 8.3.425 and was caused by 02e16b8bce8f9d19fef32932a4742e797025cc19.

I am already telling the configure script where my version 3.x python executable is located, e.g.:

```
./configure ac_cv_path_PYTHON3=/opt/local/bin/python3.11
```

but this makedbh script is ignoring that.

donn:
* Adds Python 3 to macOS requirements
* Gate include of `DEPEND_FILE` in `rules.mak` on whether the target is `clean` or not (the include isn't pre-processed; it will attempt to build everything in the include otherwise)
* Make `makedbh` a configurable file- the shebang now uses the discovered path for Python3

---
Resolves #265

### DIFF
--- a/INSTALL_MacOS.md
+++ b/INSTALL_MacOS.md
@@ -3,7 +3,7 @@
 Get [Homebrew](https://brew.sh).
 
 ```sh
-brew install cairo tcl-tk
+brew install cairo tcl-tk python3
 brew install --cask xquartz
 ./scripts/configure_mac
 make database/database.h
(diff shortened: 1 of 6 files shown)


RTimothyEdwards (approved):
Looks like that should do the trick.

RTimothyEdwards:
Pulled and merged, thanks!

PR #282, opened 2023-12-04
More explicit pointer truthiness check

donn:
`magic/database/DBio.c::DBOpenOnly` calls `dbReadOpen(CellDef *, bool, bool, int)` with a pointer for the second argument. While technically valid C, newer compilers (Clang 16) have recently frowned upon implicit pointer-to-int conversion, requiring the flag `-Wno-int-conversion` or else an error is raised.

Given that there is one such instance in the codebase, I think fixing it would be the simplest solution.


RTimothyEdwards (approved):
Looks good to me.

PR #283, opened 2023-12-10
Revamped toolbar

Landflier:
Includes the new toolbar, with all previous functionalities working. Also included is the reorderToolbar script, which is under development.


RTimothyEdwards (approved):
Looks good to me;  will test from my end.

RTimothyEdwards:
Pulled and merged in opencircuitdesign.com.

PR #288, opened 2024-01-23
Use raw strings in preproc.py

ryandesign:
When Python 3.12 is used to build magic, these warnings appear:

```
magic/../scripts/preproc.py:62: SyntaxWarning: invalid escape sequence '\('
  defrex = re.compile('defined[ \t]*\(([^\)]+)\)')
magic/../scripts/preproc.py:63: SyntaxWarning: invalid escape sequence '\|'
  orrex = re.compile('(.+)\|\|(.+)')
magic/../scripts/preproc.py:65: SyntaxWarning: invalid escape sequence '\|'
  notrex = re.compile('!([^&\|]+)')
magic/../scripts/preproc.py:66: SyntaxWarning: invalid escape sequence '\('
  parenrex = re.compile('\(([^\)]+)\)')
magic/../scripts/preproc.py:170: SyntaxWarning: invalid escape sequence '\('
  paramrex = re.compile('^([^\(]+)\(([^\)]+)\)')
magic/../scripts/preproc.py:180: SyntaxWarning: invalid escape sequence '\*'
  ccstartrex = re.compile('/\*')                # C-style comment start
magic/../scripts/preproc.py:181: SyntaxWarning: invalid escape sequence '\*'
  ccendrex = re.compile('\*/')          # C-style comment end
magic/../scripts/preproc.py:417: SyntaxWarning: invalid escape sequence '\('
  pcondition = condition + '\('
magic/../scripts/preproc.py:420: SyntaxWarning: invalid escape sequence '\)'
  pcondition += '(.*)\)'
magic/../scripts/preproc.py:426: SyntaxWarning: invalid escape sequence '\g'
  pvalue = pvalue.replace(param, '\g<' + str(idx) + '>')
```

I've fixed it by using raw strings. See https://docs.python.org/3/library/re.html:

> The solution is to use Python’s raw string notation for regular expression patterns; backslashes are not handled in any special way in a string literal prefixed with `'r'`. So `r"\n"` is a two-character string containing `'\'` and `'n'`, while `"\n"` is a one-character string containing a newline. Usually patterns will be expressed in Python code using this raw string notation.

I've run the build four times: with Python 3.11 (with and without my patch) and Python 3.12 (with and without my patch). The output of preproc.py (proto.magicrc) is identical in all four runs, but with my patch Python 3.12 no longer produces warnings. And presumably in some future version of Python the warning will become an error.

### DIFF
--- a/scripts/preproc.py
+++ b/scripts/preproc.py
@@ -59,13 +59,13 @@ import sys
 
 def solve_statement(condition):
 
-    defrex = re.compile('defined[ \t]*\(([^\)]+)\)')
-    orrex = re.compile('(.+)\|\|(.+)')
-    andrex = re.compile('(.+)&&(.+)')
-    notrex = re.compile('!([^&\|]+)')
-    parenrex = re.compile('\(([^\)]+)\)')
-    leadspacerex = re.compile('^[ \t]+(.*)')
-    endspacerex = re.compile('(.*)[ \t]+$')
+    defrex = re.compile(r'defined[ \t]*\(([^\)]+)\)')
+    orrex = re.compile(r'(.+)\|\|(.+)')
+    andrex = re.compile(r'(.+)&&(.+)')
+    notrex = re.compile(r'!([^&\|]+)')
+    parenrex = re.compile(r'\(([^\)]+)\)')
+    leadspacerex = re.compile(r'^[ \t]+(.*)')
+    endspacerex = re.compile(r'(.*)[ \t]+$')
 
     matchfound = True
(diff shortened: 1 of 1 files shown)


RTimothyEdwards (approved):
It's very disappointing to think that the Python developers would just redefine the operation of a core language feature that would make it backwardly-incompatible with uncountable numbers of existing scripts.  But thanks for the patch!

RTimothyEdwards:
Pulled and merged into version 8.3.460 on opencircuitdesign.com.

PR #296, opened 2024-03-28
Fix select more|less processing

### LINKED ISSUE #295: select more cell <instance> does not work.
magic version 8.3.465

Opening a hierarchical layout and executing `cellname childinst` will give a list of use instance names.

`select cell <instance_name>` will select one cell, but `select more cell <instance_name>` gives the following error
```
Couldn't find a cell use named "at"
```

This is caused by the argument count being decremented (for example, after processing `more` or `less`), but not accounted for when adding a coordinate value.
For example, in `commands/CmdRS.c` `tx_argc` is decremented here
```
        if (!strncmp(cmd->tx_argv[1], "more", arg1len))
        {
            more = TRUE;
            less = FALSE;
            optionArgs = &cmd->tx_argv[2];
            cmd->tx_argc--;
        }
```
but when coordinates added to original command array, since `tx_argc` hasa been decremented, the last value will be overwritten with `at`.
```
            {
                char *aptr;
                int i;

                window = CmdGetRootPoint((Point *) NULL, &scx.scx_area);

                /* Recast command with "at x y" at the end for logging */
                for (i = 0; i < cmd->tx_argc; i++)
                {
                    aptr = cmd->tx_argv[i] + strlen(cmd->tx_argv[i]);
                    *aptr = ' ';
                }
                sprintf(aptr + 1, "at %di %di", scx.scx_area.r_xbot,
                                scx.scx_area.r_ybot);
                TxRebuildCommand(cmd);

            }
```
Pull request will be posted shortly.

d-m-bailey:
Fixes #295 

However, I feel that this is a bad patch to a poor implementation.

### DIFF
--- a/commands/CmdRS.c
+++ b/commands/CmdRS.c
@@ -859,6 +859,7 @@ CmdSelect(w, cmd)
     char *tclstr;
     Tcl_Obj *lobj;
 #endif
+    int original_cmd_length = cmd->tx_argc;
 
 /* How close two clicks must be to be considered the same point: */
 
@@ -1317,7 +1318,8 @@ Okay:
 	        window = CmdGetRootPoint((Point *) NULL, &scx.scx_area);
 
 		/* Recast command with "at x y" at the end for logging */
-		for (i = 0; i < cmd->tx_argc; i++)
+		int argc_adjustment = original_cmd_length - cmd->tx_argc; // need to restore argc adjustments;
+		for (i = 0; i < original_cmd_length; i++)
 		{
 		    aptr = cmd->tx_argv[i] + strlen(cmd->tx_argv[i]);
 		    *aptr = ' ';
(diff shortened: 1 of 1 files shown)


d-m-bailey:
There may be a better solution. Don't merge this just yet.

d-m-bailey:
@RTimothyEdwards Instead of modifying `tx_argc`, I was considering using a `pstart` variable like in `commands/CmdFI.c` to keep track of the first relevant parameter. However, `tx_argc` is used in many places.
In other `commands/Cmd*.c` programs, `locargc` is used.

The solution in this merge is a kludge, but it may be a minimal change. 

Could you please decide what to do?

RTimothyEdwards:
I just pushed what I think is the right solution to this issue.  The CmdSelect() code was messing with the number of arguments directly instead of keeping a separate count of non-qualifier arguments.  My correction appears to solve the issue with "select more cell" and "select less cell" without breaking anything else, and the code is cleaner.

d-m-bailey:
@RTimothyEdwards as always, thanks for looking into this. I'll test it once it's on the public repo.

RTimothyEdwards:
I have not had any issues since the fix, so I will close this now.

PR #314, opened 2024-06-04
3 x crash fix and 1 x read uninitialized fix

dlmiles:
The `findUse()` probably wants a better review to confirm the UNDO operation left in place is enough, in particular DBundo.c:812

### DIFF
--- a/calma/CalmaWrite.c
+++ b/calma/CalmaWrite.c
@@ -1440,7 +1440,7 @@ calmaOutFunc(def, f, cliprect)
 		numports++;
 	    }
 	}
-	if (newll != NULL)
+	if (ll != NULL)
 	{
 	    /* Turn linked list into an array, then run qsort on it	*/
 	    /* to sort by port number.					*/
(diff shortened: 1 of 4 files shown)


dlmiles (commented):
Maybe the real fix this should be `if (ll != NULL)`  ???   camlaWrite.c:1443

dlmiles:
I took my own suggestion after review in comment https://github.com/RTimothyEdwards/magic/pull/314#pullrequestreview-2096688410 

Also looking at the lef/lefWrite.c there was another nearby site which looks like it could trigger the same problem, so I put in the same if() guard.  I can also see the same check exists already around line 1560.

There is one more site around line 1723 that is unclear if there is a risk there also.

RTimothyEdwards (approved):
All of this looks good to me.
I do wish you would put a space after "if", though . . . "if(x)" is not a function call.

RTimothyEdwards:
Pulled and merged.

PR #316, opened 2024-07-30
Fix infinite def write

kareefardi:
I think `.def_part` is resolved to `.def` resulting in infinite write behavior. This renames the partial file `.._part.def` avoiding the infinite loop while writing def.

I am not sure however if `freeMagic(outName_part)` is called in all the correct places.

### DIFF
--- a/lef/defWrite.c
+++ b/lef/defWrite.c
@@ -2966,6 +2966,13 @@ DefWriteCell(def, outName, allSpecial, units, analRetentive)
 	TxError("Please name the cell before generating DEF.\n");
 	return;
     }
+    char * outName_part ;
+    if((outName_part = mallocMagic(strlen(outName)+strlen("_part")+1)) != NULL) {
+        outName_part[0] = '\0';
+        strcat(outName_part, outName);
+        strcat(outName_part, "_part");
+    }
+    //
 
     f = lefFileOpen(def, outName, ".def", "w", &filename);
 
@@ -3016,7 +3023,7 @@ DefWriteCell(def, outName, allSpecial, units, analRetentive)
 
     /* Not done yet with output, so keep this file open. . . */
 
-    f2 = lefFileOpen(def, outName, ".def_part", "w", &filename);
+    f2 = lefFileOpen(def, outName_part, ".def", "w", &filename);
 
     if (f2 == NULL)
     {
@@ -3030,7 +3037,8 @@ DefWriteCell(def, outName, allSpecial, units, analRetentive)
 	/* If part 2 cannot be opened, remove part 1 */
 	fclose(f);
 	unlink(filename1);
-        freeMagic(filename1);
+    freeMagic(filename1);
+    freeMagic(outName_part);
 	return;
     }
     filename2 = StrDup((char **)NULL, filename);
@@ -3133,7 +3141,7 @@ DefWriteCell(def, outName, allSpecial, units, analRetentive)
 
     /* Append contents of file with NETS and SPECIALNETS sections */
 
-    f2 = lefFileOpen(def, outName, ".def_part", "r", &filename);
+    f2 = lefFileOpen(def, outName_part, ".def", "r", &filename);
     if (f2 == NULL)
     {
 	/* This should not happen because the file was just written. . . */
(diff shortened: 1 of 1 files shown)


lef/defWrite.c
dlmiles, line 1: Consider this slightl tweak to approach, `outName` can be NULL (according to function docs), the lefFileOpen() will return the actual filename used (after resolving what it will be, when not provided via outName).  So you can't do things like strlen(NULL).

The codebase doesn't really have NULL checks on failed mallocMagic() calls, so this mitigated to ASSERT and removes the whole what to do when it fails consideration the if() block implies.

I'm assuming the `filename` contains the file extension (.def/.lef) as it is used directly with `unlink()` as-is.

So this moves the block to be lower down around line 2988 of the original file.

```
    filename1 = StrDup((char **)NULL, filename); // from original file line 2985

    char *outName_part;
    {   /* use the 'filename' given back from lefFileOpen() */
        outName_part = mallocMagic(strlen(filename)+strlen("_part")+1);
        ASSERT(outName_part, "outName_part");
        strcpy(outName_part, filename);
        char *cp = strrchr(outName_part, '.');
        if (cp)
           *cp = '\0'; /* remove extn */
        strcat(outName_part, "_part");
    }

    defWriteHeader(def, f, scale, units); // from original file line 2987 (remove this comment)
```

RTimothyEdwards:
@dlmiles :  My problem with the original pull request is that I still don't understand the nature of the infinite loop and I don't see how the code change fixes it, and I don't understand the need to change the file name being written out.  This is mostly do to me having limited time to check and test code changes.

dlmiles:
Understand, just trying to help out with the code review aspect to point out `strlen(outName)` is problematic because outName can be NULL.  The returned variable `filename` looks a better variable to use.

I am not affected by the problem this commit is trying to address.

@kareefardi is the infinite loop an internal magic infinite loop?  Or an issue with external tooling seeing or not seeing well formed files with the appropriate file extension?

Is there an example command sequence to run with project files to demonstrate the original problem?

kareefardi:
@dlmiles @RTimothyEdwards I was able to isolate the problem to it being caused by calling `def write spm.def` instead of `def write spm`. Here is a demonstration of this: 
demo

dlmiles:
```
diff --git a/lef/lefWrite.c b/lef/lefWrite.c
index 024fdf2a..639cef78 100644
--- a/lef/lefWrite.c
+++ b/lef/lefWrite.c
@@ -191,10 +191,12 @@ lefFileOpen(def, file, suffix, mode, prealfile)
     {
        if (strcmp(endp, suffix))
        {
+#if 0
            /* Try once as-is, with the given extension.  That takes care   */
            /* of some less-usual extensions like ".tlef".                  */
            if ((rfile = PaOpen(name, mode, NULL, Path, CellLibPath, prealfile)) != NULL)
                return rfile;
+#endif
 
            len = endp - name;
            if (len > sizeof namebuf - 1) len = sizeof namebuf - 1;

```

This tries a re-open of the given files "spm.def" and this is a hit and work, so it returns a handle.
This is due to the suffix=".def_part" not matching the "spm.def" extension.

dlmiles:
Order of events:

```
DefWriteCell(outName="spm.def") // or outName="spm" which does not trigger issue
f = lefFileOpen(def, outName, ".def", "w", &filename);
## 'f' has data written into file

f2 = lefFileOpen(def, outName, ".def_part", "w", &filename);
## data is written into 'f2'  this is acts like a tmpfile I guess due to out-of-order data processing
## but it is in the same directory as the outName
## it is unclear (and maybe not important) if this file actually unique and gets created, it maybe truncating 'f' but both files have data written into them now

## f2 has data written into it
fclose(f2);

## f1 has more data written into it (call it 512M of data)

f2 = lefFileOpen(def, outName, ".def_part", "r", &filename);
## f2 is reopening the previously processed data, but this reopen actually opens the same file as 'f' not a seperate

## at this point:
## 'f2' is RO handle on 'spm.def' at filepos=0
## 'f'is RW handle on 'spm.def' at filepos=512M
while (fgets(line, sizeof line, f2) != NULL) fprintf(f, "%s", line);
fclose(f2);
## this while loop causes infinite write, as the file handles are 512M apart of the same file
```

The optimization (if that is a good term) in the `#if 0` commented out section, uses the outName as-is and has suffix=NULL in call to PaOpen() so it ignores the passed suffix=".def_part" passed to lefFileOpen() so when the optimization is triggered it opens `spm.def`

Previously (before this PR) outName is the same value to all calls to lefFileOpen(), which then derives the `endp` (extension) from that it can see in outName.  endp=".def"  suffix=".def_part"

By modifying outName to append `_def_part`for the tempfile it permutes the filename enough so it can not match the 'f' filehandle filename of `spm.def`

RTimothyEdwards:
Fixed to my satisfaction in PR #359 , so I will now close this PR.

PR #325, opened 2024-09-30
RFC: 102 x CodeQL/SonarCloud/Copilot Autofix guided patches (high/medium severity concerns)

dlmiles:
Request For Comments

IMHO There are few false positives here, over two-thirds look like important fixes worth bringing to attention.
Issues are in the High and Medium categories of the respective tooling.
This was an opportunity to try out the 3 technologies on a sufficiently complex project.

The solutions were timeboxed ("is it possible to fix the concern in 5 minutes of coding and 5 minutes of administration?"  on average).  A number of more complex issues still exists (maybe difficult to trigger concern, due to setup/scenario complexity).  This patch set contains the most serious as categorized by the tooling.

Some areas like the multiplication promotion to 'double' from 'float' before multiplication operation are expected to have a visible effect on emitted and generated files.  I'm looking for a way to run 100s of projects through magic to compare and validate data processing aspects.

Please advise if the code analysis tooling links to not work publicly, as the intention is the information is public in line with the open source nature of the project.

Please advise if there is any path to proceeding with the outcome, maybe only a sub-set of commits are taken, maybe some need rework, etc...  alternatively feel free to cherry-pick on particular issues and report back if they solve your problem (or create new ones).

### DIFF
--- a/calma/CalmaRead.c
+++ b/calma/CalmaRead.c
@@ -345,7 +345,7 @@ done:
 bool
 calmaParseUnits()
 {
-    int nbytes, rtype;
+    int nbytes, rtype = 0;
     double metersPerDBUnit;
     double userUnitsPerDBUnit;
     double cuPerDBUnit;
--- a/commands/CmdWizard.c
+++ b/commands/CmdWizard.c
@@ -24,6 +24,7 @@ static char rcsid[] __attribute__ ((unused)) = "$Header: /usr/cvsroot/magic-8.0/
 #endif  /* not lint */
 
 #include <stdio.h>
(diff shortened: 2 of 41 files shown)


RTimothyEdwards (approved):
Impressive.  Quite impressive.  Especially how thorough this is, checking obscure parts of the code (like some of the old routers) that I never touch.  I have reviewed all of the changes, and they all appear to be proper.  I would be interested in seeing what this method produces if not limited to quickly-validated changes.

RTimothyEdwards:
Pulled and merged.

PR #326, opened 2024-10-01
database/DBio.c: Fix MAGIC_WRAPPER #endif to fix CI WASM build

dlmiles:
Misplaced #endif causing the build to break when option enabled error introduced in 3dc5018

Related commits:

commit 68a088943f5534ac356ce8113d3985868b35d03e
Date:   Tue Sep 12 11:12:00 2023 -0400

commit 3dc5018af4994cc23e2c2dbc822927d7897fba93 (tag: 8.3.477)
Date:   Fri May 3 21:43:27 2024 -0400

### DIFF
--- a/database/DBio.c
+++ b/database/DBio.c
@@ -688,8 +688,8 @@ dbCellReadDef(f, cellDef, ignoreTech, dereference)
 			    freeMagic(argv[1]);
 			    freeMagic(argv[0]);
 			}
-#endif
 		    }
+#endif
 		}
 		if (strcmp(DBTechName, tech))
 		{


RTimothyEdwards (approved):
Good catch, although if you ran into this error, then you were compiling magic without Tcl/Tk support, which results in what is, effectively, crippleware.  However, you may have been testing all compile-time options.

RTimothyEdwards:
Pulled and merged.  Will update the version number after reviewing and handling some other pull requests.

dlmiles:
> Good catch, although if you ran into this error, then you were compiling magic without Tcl/Tk support, which results in what is, effectively, crippleware.  However, you may have been testing all compile-time options.

The configuration is the GitHub Action WASM build.  I was diagnosing why it is showing failure to understand what changed.

PR #327, opened 2024-10-01
Github Action Maintenance update

dlmiles:
Should remove CI deprecation warnings.
Deduplicate the workflow names (currently 2 exist both called CI, this renames one of them)
Use --disable-compression for WASM build

### DIFF
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -3,7 +3,7 @@ on:
     tags:
       - "*"
 
-name: CI
+name: CI-appimage
 
 jobs:
   build_appimage:
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
       - name: Get the version
         id: get_version
         run: |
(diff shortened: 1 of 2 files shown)


RTimothyEdwards:
What is the point of using "--disable-compression"?

dlmiles:
> What is the point of using "--disable-compression"? 

My gut tells me if it ever did find the libz library (when `configure` searches system libraries) the final WASM build either would not link or does not include it anyway. 

If the WASM binary is to have support for compression the build needs to compile libz from source to target WASM then it will link and have the feature available. 

I can this theory later to confirm.

dlmiles:
My local test building has `apt install zlib1g lib1g-dev` installed (standard system C ABI libraries), but the WASM compiler options during a WASM build do not have the `-DHAVE_ZLIB=1` present.
So in effect the feature is not available.  So the purpose of adding the option to the GHA .yml is to convey to the reader the feature is not expected to work or be available.


Also I note with the standard Linux build in CI maybe the following packages should be added `apt-get install zlib1g-dev libgl-dev libglu1-mesa-dev` which is based on Ubuntu 24.04 LTS.
This will better auto-detect all features and cover building of more lines of code with most features enabled ?  So the CI is giving better feedback.

RTimothyEdwards:
@dlmiles :  Good point;  I didn't realize that I had not updated my install instructions to include zlib after adding the feature to read and write compressed GDS files.

RTimothyEdwards (approved):
Looks good to me.  Will merge after I finish debugging something.

RTimothyEdwards:
Pulled and merged.

PR #332, opened 2024-10-05
K&R calma/**

dlmiles:
This series converts the calma module into ANSI function prototype syntax.

Please feedback if the method and commit breakdown helps the review process (for example github has single character change highlight, the code style adopted is to help get through review, a future commit could make just style cleanup/white-space only change which can be reviewed better via diff -b)

There are 2 subsequent change set behind this one (calma/** .rodata cleanup, ATTR_FORMAT_PRINTF - reminder to self)

### DIFF
--- a/calma/CalmaRdcl.c
+++ b/calma/CalmaRdcl.c
@@ -96,8 +96,8 @@ typedef struct {
  */
 
 OFFTYPE
-calmaSetPosition(sname)
-    char *sname;
+calmaSetPosition(
+    char *sname)
 {
     OFFTYPE originalPos = 0, currentPos = 0;
     int nbytes, rtype;
@@ -188,7 +188,7 @@ calmaSetPosition(sname)
  */
 
 void
-calmaNextCell()
+calmaNextCell(void)
 {
     int nbytes, rtype;
 
@@ -240,7 +240,7 @@ calmaNextCell()
  */
 
 Plane **
-calmaExact()
+calmaExact(void)
 {
     int pNum;
     Plane *newplane;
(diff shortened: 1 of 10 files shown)


RTimothyEdwards:
Is this supposed to be a preview change for converting all of the subroutines to have ANSI function prototypes?  Am I supposed to merge this one or just merge them all at once?

dlmiles:
Yes this can merge, if there are no major or minor objections.

If its in a PR and it builds ok, and there is no quality warning in PR comment, assume it is ready.  GH has a draft PR function to try to prevent not-ready change-sets from being merged, but still allow review to occur.

I try to make the PR a complete merge-able standalone unit.

The 2 change sets behind this one means I have not queued 2 other PRs up on Github because they have an order dependency on this PR (because the same lines or nearby lines are changed, for another reason).  They are temporarily withheld to prevent the headache of merge conflicts.

The K&R will be a slow burning matter over time, so expect a bunch, then a gap, then a bunch, etc..  i've a few more directories in alphabetical order then knee deep into 'database' but I think that will end up last as the callback cleanups are quite expansive and need a few cycles of review (on my side) to be ready.

RTimothyEdwards:
I think it's best if we don't make too many changes at a time.  Finish up with the various other issues that don't have to do with ANSI declarations, and when that seems to have settled down, we can start applying these.

RTimothyEdwards (approved):
I apologize for punting on this for three months while I worked on a critical chip tape-out.  Getting back to merging these PRs now.  This one looks fine.

RTimothyEdwards:
Pulled and merged into code base at opencircuitdesign.com.

PR #333, opened 2024-10-05
6 x fixes (hidden behind compiler warnings)

dlmiles:
All commits are untested and unchecked.  The fix is based on evaluation of the context and code around to try and GUESS the original programming intention.

These changes are separated from other commit due to requiring a more detailed inspection as they will all change magic computation behaviour.


RTimothyEdwards (approved):
I have tried to check these as carefully as I can, and as far as I can tell, they are all valid and the modifications are correct.

RTimothyEdwards:
Pulled and merged.

PR #334, opened 2024-10-05
GCC14 -Wall cleanup series

dlmiles:
This cleans up all warnings that have fixes that should not change application behaviour.


dlmiles:
Correction to DBio.c in one of the commits.

  DBio.c edited as my original change was caught out by the incorrect
  indention of the 'else' keyword and my modification would have caused a
  NULL deref.   Exactly the kind of issue this warning is trying to prevent.

drc/DRCtech.c
```diff
@@ -4138,7 +4138,7 @@ DRCTechScale(scalen, scaled)
     int scalegcf;
 
     if (DRCCurStyle == NULL) return;
-    else if (scalen == scaled == 1) return;
+    else if (scalen == scaled) return;
```
dlmiles, line 4141: This needs integer truth table test ranges -1,0,1,2 the boolean table example in commit may not be enough. 

a == b == 1

Maybe:
a == 1 && b == 1 (also a == b && b == 1)
Or maybe:
a !=0 && b == 1

These are not quite the same thing.

RTimothyEdwards: The scale factor is N/D (numerator / denominator), so any N/D equal to 1 means that DRCTechScale() will have no effect and does not need to be run.  So your original correction appears to be right.

RTimothyEdwards (approved):
I always find this sort of syntax enforcement annoying;  "if (a = b)" has always looked quite obvious to me;  however, if simply doing "if ((a = b))" makes the compiler happy, then I guess its just a style change.

I can't swear I started at every correction, but I think so.  I looked carefully at a handful that were not so obvious, and they all looked okay to me.

RTimothyEdwards:
Pulled and merged.

dlmiles:
Agree modification is same as original `(a == b)`
```
a   b    old new
-1, -1 =  1  1 SAME
-1,  0 =  0  0 SAME
-1,  1 =  0  0 SAME
-1,  2 =  0  0 SAME
 0, -1 =  0  0 SAME
 0,  0 =  1  1 SAME
 0,  1 =  0  0 SAME
 0,  2 =  0  0 SAME
 1, -1 =  0  0 SAME
 1,  0 =  0  0 SAME
 1,  1 =  1  1 SAME
 1,  2 =  0  0 SAME
 2, -1 =  0  0 SAME
 2,  0 =  0  0 SAME
 2,  1 =  0  0 SAME
 2,  2 =  1  1 SAME
```

PR #336, opened 2024-10-06
clang18 default warnings and build fix

dlmiles:
These changes not modify application behaviour, they address clang18 warnings.

one commit fixes building with clang18 concerning etext symbol use.

clang18 was also then put into a stricter standards compliant mode (for some of the concerns) to help highlight issues.


RTimothyEdwards (approved):
The propertyvalue fix in DBio.c should really be to keep propertyvalue[2048] and change everything else to match it, since no compiler is going to allocate exactly 2049 bytes.  But this is the quick and easy solution.

RTimothyEdwards:
Pulled and merged.

PR #340, opened 2024-10-09
Added CmdFlush_NoConfirm option to disable flush prompt

wulffern:
In commands/CmdFI.c.

To ignore prompt
```tcl
set CmdFlush_NoConfirm 1
```

To enable prompt again
```tcl
unset CmdFlush_NoConfirm 
```

### DIFF
--- a/commands/CmdFI.c
+++ b/commands/CmdFI.c
@@ -943,6 +943,7 @@ CmdFlush(w, cmd)
     static char *actionNames[] = { "no", "yes", 0 };
     char *prompt;
     bool dereference = FALSE;
+    char *strNoConfirm;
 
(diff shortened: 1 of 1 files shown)


RTimothyEdwards (commented):
This breaks conventions.  I would rather see it implemented as command option `flush -noprompt` which is the convention used for the `quit` command as well.  That will require changing the argument parsing to use `Lookup()` rather than the simple string compare (which also, _mea culpa_, breaks convention)

wulffern:
Sure thing, I'll give it a crack. To be continued.

dlmiles:
issue#341 link issue

wulffern:
I'm closing this PR as it's fixed by PR #341

PR #341, opened 2024-10-09
PR#340 alternative ?  flush -noconfirm .. exit exit_status -noconfirm

dlmiles:
@wulffern please advise, test as alternative to PR#340

Documented hidden/secret `exit [-noconfirm]` option to disable exit prompt
Added `exit [exit_status]` option to send exit_status number
Added `flush -noconfirm` option to disable flush prompt

I have tested the 'exit' patches.

I have not yet tested the 'flush' patch.  Still comment back when testing complete on my side but would appreciate feedback if it closes your concern.

### DIFF
--- a/commands/CmdFI.c
+++ b/commands/CmdFI.c
@@ -943,6 +943,7 @@ CmdFlush(w, cmd)
     static char *actionNames[] = { "no", "yes", 0 };
     char *prompt;
     bool dereference = FALSE;
+    bool noprompt = FALSE; /* no interactive confirm when changed */
 
     if (!strncmp(cmd->tx_argv[cmd->tx_argc - 1], "-deref", 6))
     {
@@ -950,9 +951,15 @@ CmdFlush(w, cmd)
 	cmd->tx_argc--;
     }
 
+    if (!strcmp(cmd->tx_argv[cmd->tx_argc - 1], "-noprompt"))
+    {
+        noprompt = TRUE;
+        cmd->tx_argc--;
+    }
+
     if (cmd->tx_argc > 2)
     {
-	TxError("Usage: flush [cellname] [dereference]\n");
+	TxError("Usage: flush [cellname] [-noprompt] [-dereference]\n");
 	return;
     }
 
@@ -973,7 +980,9 @@ CmdFlush(w, cmd)
 	}
     }
 
-    if (def->cd_flags & (CDMODIFIED|CDSTAMPSCHANGED|CDBOXESCHANGED))
+    bool has_changes = (def->cd_flags & (CDMODIFIED|CDSTAMPSCHANGED|CDBOXESCHANGED)) != 0;
+
+    if (!noprompt && has_changes)
     {
 	prompt = TxPrintString("Really throw away all changes made"
 			" to cell %s? ", def->cd_name);
@@ -984,7 +993,7 @@ CmdFlush(w, cmd)
 
     cmdFlushCell(def, dereference);
     SelectClear();
-    TxPrintf("[Flushed]\n");
+    TxPrintf("[Flushed%s]\n", has_changes ? " Modifications were Discarded" : "");
 }
 
 
--- a/doc/html/flush.html
+++ b/doc/html/flush.html
@@ -26,7 +26,7 @@ last saved version.
 
 <H3>Usage:</H3>
    <BLOCKQUOTE>
-      <B>flush</B> [<I>cellname</I>] [<B>-dereference</B>]<BR><BR>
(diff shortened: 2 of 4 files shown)


RTimothyEdwards (approved):
Using `strncmp()` for testing each command argument forces the arguments to be in a specific order.  The preferable method is to use an option structure with the `Lookup()` routine, as is done for other commands with multiple options.  Yes, I know I started the bad behavior. . .

wulffern:
@dlmiles Yup, I can confirm that the flush -noprompt does what I want. Thanks.

It's a bit more finicky to use since I need to know that the -noprompt option is in the Magic version. 

Magic throws an error if I use -noprompt in a Magic version that does not have it, but with a "flush_all" script like the below it works fine.

```tcl
set values [ cellname list all ]
set mrev [lindex [lreverse [split [exec magic --version] "."]] 0 ]
foreach x $values {
    if {${mrev} > 495} {
        flush $x -noprompt
    } else {
        flush $x
    }
}
```

PS: I could not find the revision in any global tcl variable, nor was I able to redirect "version" to a variable, but I'm sure there is a more elegant way to do the above.

RTimothyEdwards:
@wulffern :  `catch {flush -noprompt}` should work around the backwards-compatibility failure, or maybe
```
if {[catch {flush -noprompt}]} {flush}
```

dlmiles:
@RTimothyEdwards  Did you want a revision using `Lookup()` ?  not clear as it PR marked approved.


@wulffern maybe you can setup $ENV{'MAGIC_VERSION'} before magic is run, to TCL inherits envars, like in a startup wrapper shell script ?  This way it is always available.  This isn't much difference to how you are doing it now really.  Maybe there should be a patch to make such a thing available from TCL an auto-populated variable.

Maybe it can be run as `flush $x -noprompt` and if this errors, too many args, it just retries the command without `-noprompt`.  This may not be a possible if `$x` is not present, as `-noprompt' maybe processed as a `cellname` which will probably also error.


Difficult to provide a backward compatible mechanism that isn't as finicky, as least you can make it into `myflush` TCL proc to clean / hide this extra work?

Your original patch with an environment variable set from inside (or maybe inherited from outside) might be better reworked with the `batch_mode` feature, such that you can start up in GUI mode with batch_mode and it has the effect of making all interactive decision dialogs to cause error termination.  It also make sense for $ENV{'MAGIC_BATCH_MODE'} to also be a thing so scripts a runtime.  Maybe this is already possible.

Which then necessitates an option like `-noprompt` for every command that might present a dialog, which allows the command to succeed.  This is my use case, I am looking to understand how some test cases maybe prepared that operate in GUI mode, but I need a way to manage dialog interactions and terminate with exit_status.  Hence this PR.

RTimothyEdwards:
@dlmiles :  I approved it because the code is usable as-is, but if you can change it to use `Lookup()` I'd appreciate it.

FYI:  You can use the `version` command in magic to get the version;  it does not need to come from an external variable.  Although a `version --list` option would be nice to get just the version numbers back as a Tcl list instead of text.

dlmiles:
I have checked over the Lookup() methods and the usages in the codebase I can't quite see an obvious way all the requirements can be met :

* allow -option to be any order (ok regular Lookup usage does this part)
* continue to ensure `-nodef` is the same as the longer form `-nodeference` but also `-nodefer` etc.. to ensure a users script doesn't break  (this is the main blocker)
* extract the `[exit_status]` option, is this allowed to be in any position ?   (unclear how the remaining arguments left in the list, unclaimed by any Lookup is extracted into a canonicalized form argv[])  so the arg count in this category can be seen, to allow `exit -noprompt 42` but also check for invalid `exit -noprompt 42 0`
* error on weird usage like `exit 1 -noprompt -noprompt -nodef -nodeference`

Re current patch having fixed order, sure I added in a way that should not break any current users scripts and documentation indicates correct order to make it as easy to discover and get working as possible.

RTimothyEdwards (approved):
Looks good;  I like the addition of the forced exit status, which will be very useful with scripted batch processes.

RTimothyEdwards:
Pulled and merged into the opencircuitdesign.com code base.

dlmiles:
Not sure what happened to this change set it did not make it into master

Instead there is 772bfe2f71f31bf8e0254fe685c8bc65e0829709 from Carsten  (and the initial version that was using a non-standard method to activate the feature)

So I rebase 772bfe2f with a revert ahead of the 3 commits here.

dlmiles:
@RTimothyEdwards  No re-open issue button here.

PR #343, opened 2024-10-10
Github Action MacOSX revamp

dlmiles:
The FREAD macro name is renamed due to clash with sys/fcntl.h in Xcode SDK on MacOSX.

The GHA will build for both legacy x86_64 and newer ARM64 platforms.

The configure_mac resolves an issue with the wrong X/Xlib.h header being included due to a (bad) copy being bundled with tk-tcl when the copy from quartz is the one needed.


dlmiles:
Updated: Somehow lost TMPFILE=value assignment in YML during final prepare

dlmiles:
Next iteration on MacOS CI in Github Acton:
* diagnostics improved
* autoconf X11 compiler/linker switches fixed (maybe this wasn't causing an issue, but manual inspection of the -L/usr/X11/lib placement indicated it would have no effect at the end of the linker options)
* Github Action Summary output (everything you might want to know is listed)
* Performs install
* Performs kick the tyres testing, to confirm it runs without crashing
* It prepares download files with what is built (maybe the binaries work, maybe they don't, I try to solicit feedback)
As before:
* Builds for x86_64 and ARM64 MacOS

RTimothyEdwards (approved):
I will need to test the effect of X_PRE_LIBS and X_EXTRA_LIBS on a non-Mac system, but it looks okay to me.

RTimothyEdwards:
Pulled and merged into opencircuitdesign.com code base.

PR #344, opened 2024-10-10
SonarCloud guided improvements

dlmiles:
These are a little more involved changes, to better understand the context and decide on a strategy.
Most commits take a defensive coding approach to scenario SonarCloud may present.

### DIFF
--- a/graphics/grOGL1.c
+++ b/graphics/grOGL1.c
@@ -527,6 +527,8 @@ pipehandler()
 
 	    entry = HashLookOnly(&grOGLWindowTable, (char *)VisEvent->window);
 	    mw = (entry)?(MagWindow *)HashGetValue(entry):0;
+	    if (!mw)
+		break;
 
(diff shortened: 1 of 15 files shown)


graphics/grOGL1.c
```diff
@@ -1076,7 +1078,8 @@ GrOGLIconUpdate(w,text)
 
      if (brack = strchr(text,'['))
      {
-     	  brack--;
+	  if (brack != text)
```
dlmiles, line 1081: Consider if this should move up into the main if() 2 lines above - note to self

dlmiles:
Resolved my own query slightly differently.  Found a total of 5 sites with similar pattern, these updated too.

RTimothyEdwards (approved):
All of these changed look correct.

RTimothyEdwards:
Pulled and merged into opencircuitdesign.com code base.

PR #345, opened 2024-10-10
Enable compiler __attribute__ assistance

dlmiles:
Draft as needs needs minor rebase after pr #332

### DIFF
--- a/calma/CalmaRead.c
+++ b/calma/CalmaRead.c
@@ -459,7 +459,7 @@ calmaParseUnits(void)
  */
 
 void
-CalmaReadError(char *format, ...)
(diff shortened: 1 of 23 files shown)


RTimothyEdwards:
I will wait on your rebase.  Sorry for the long delay.

RTimothyEdwards (approved):
Looks good to me.  I'm not too familiar with compiler attributes.  It looks like that would be very helpful in removing some of the ugly in-lined definitions to make them subroutines with an "inline" attribute.  But that's a task for another day.

RTimothyEdwards:
Pulled and merged into opencircuitdesign.com repository..

PR #347, opened 2024-10-11
GHA: canary-matrix.yml

dlmiles:
This adds a canary build to the CI that will build against old/new linux to provide an indication using various build options to confirm how healthy the source tree and release is.

It is setup to run on tagging (releases) and manually.  This should help identify if/when the build breaks when using bleeding edge versions of Linux systems, the diagnostic output should help provide details of what thing changed to cause that.

Output example can be seen at https://github.com/dlmiles/magic/actions/runs/11283759998

### DIFF
--- a/.github/workflows/canary-matrix.yml
+++ b/.github/workflows/canary-matrix.yml
@@ -1,0 +1,285 @@
+
+name: CI-canary-matrix
+
+on:
+  push:
+    tags:
+      - "*"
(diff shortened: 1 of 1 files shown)


RTimothyEdwards (approved):
No issues with this.

RTimothyEdwards:
Pulled and merged in the opencircuitdesign.com code base.

PR #351, opened 2024-10-21
TCL9 support

dlmiles:
Initial attempt at TCL9 support.  It builds with 8.6 and 9.0 and appears to run ok, re batch script.

The most recent 4 commits still need work and resolution
* tkcon.tcl was butchered, removal of `::tk::unsupported` items, needs impact assessment
* use of `$::tk_platform` (read from migration notes)
* better ClientData handling in magic.h with TkTcl and non-TkTcl builds, ideally the tcl.h version should always have first opportunity to define this type, but right now magic.h appears to go first even in TkTcl enabled builds.
* no time has been spent on checking Tk/X11 UI is fully working
* more checking of `tcltk/*.tcl` scripts, I could not find any use of octal strings of tilde $HOME expansion

Due to the above drafting status only.

Example command sequence to build and pickup as alternative version

```
tar -zxvf tcl9.0.0-src.tar.gz
cd tcl9.0.0/unix
./configure --enable-symbols --prefix=/opt/tktcl9
make install
    
tar -zxvf tk9.0.0-src.tar.gz
tk9.0.0/unix
./configure --enable-symbols --prefix=/opt/tktcl9 --with-tcl=/opt/tktcl9/lib
make install
    
cd magic
./configure --with-tk=/opt/tktcl9/lib --with-tcl=/opt/tktcl9/lib
```


RTimothyEdwards:
I think it's best if you rebase this and then I'll check it.

dlmiles:
Rebased onto recent merge, no advancement in the outstanding points at top of issue.
Should be good enough to build successfully and take for a spin.

RTimothyEdwards (approved):
This all looks okay to me.  It better handles a number of things that I worked through to get the Tcl/Tk interpreter working, through several versions of Tcl.  I do not think that the code changes to tkcon.tcl will have any impact.  Making sure that batch scripts for magic that set an error code on exit still work is probably the highest priority.

RTimothyEdwards:
I avoided saying "things that I worked through without really knowing what I was doing" but that's pretty much the gist of it. . .

RTimothyEdwards:
I checked the Tk/X11 graphics interface and didn't find any issues.

Issue #322, opened 2024-08-27
-O2 compilation causes the Match function in utils/Match.c to match partial string.


d-m-bailey:
This issue is just for informational purposes.

Compiling with `-O2` optimization causes the pattern string to match any string beginning with the pattern string.
`astring` will match `astring_and_anything_else`.

Don't use `-O2`.

dlmiles:
Unable to reproduce.

Please advise the platform, distribution (/etc/os-release) and compiler (gcc -v) details.

dlmiles:
[match_main.c.txt](https://github.com/user-attachments/files/17271115/match_main.c.txt)

checkout, then configure magic project
then use command line at of file to compile testcase

please edit test case to demonstrate the claim in this bug and report platform details

d-m-bailey:
@dlmiles Well, I can't get it to work (by work I mean fail) with the snippet of code you provided. I'll see if I can recreate the problem with the original program.

PR #352, opened 2024-10-21
K&R utils/** and const-correctness (part 1)

dlmiles:
The K&R conversion of utils/** warranted a little more attention to put in place const-correctness as the same time, due to the ripple improvement effect across all the other modules consuming them.

Attention on the HEAD~5 fix commits and the 2 commits relating to PaCheckCompressed()

The commit relating to WindGetCommandTable maybe a little more tricky as it manages to non-const creation of dynamic data to then handoff for use with Lookup family functions.

The rest should be pretty much additional of 'const' keyword.
Showing as -Wall clean on older GCC8.


RTimothyEdwards (approved):
This looks okay to me, including the two items in question.

RTimothyEdwards:
Would this PR resolve Mitch's issue in #322, possibly?

RTimothyEdwards:
Merged and pulled.  Had to resolve a few conflicts caused by earlier PRs;  all conflicts appeared to be straightforward to fix.

Resolved the issue in irouter/irCommand.c by changing SetNoisyDI() to SetNoisyInt();  The SetNoisyDI() function appears to have been used in error since "mp_bloomLimit" is type int and not type dlong.

Slightly modified the code change in utils/hash.c so that the assignment to (const void *p) is done inside the "if (killFn)" block, as this avoids the additional computation in the default case where "killFn" is NULL.

Otherwise, the PR has been pulled and merged into the opencircuitdesign.com code base.

dlmiles:
> Slightly modified the code change in utils/hash.c so that the assignment to (const void *p) is done inside the "if (killFn)" block, as this avoids the additional computation in the default case where "killFn" is NULL.

Comment left at https://github.com/RTimothyEdwards/magic/commit/8371d797ade89c0a977b6e8fd9bbd35ab7cf4e2c#diff-60cb665238705d8f076fffb2132cff13b6c1ebfcd9d40b2916b353aaef406dd9R699

Lets see if the link works.  Concerning the 'h' dereference after free().

RTimothyEdwards:
It should work, because of the one-delayed free.

dlmiles:
> Would this PR resolve Mitch's issue in #322, possibly?

The changes do not address anything in #322.

My hunch is the cause has been misidentified as a compiler optimization issue.  However the exact build system, compiler version, etc.. has not been clarified so the compiler optimization claim is looking thin.  It is probably either a usage bug/error (invalid input data) or some other memory corruption matter that will hopefully become less common as improvements progress.

I have provided a profile capable testcase over there that can test any match.o generated from a buggy compiler and validate is works as expected (without the claimant needing to disclose the target project or other data).

PR #357, opened 2025-01-03
GHA: MacOS maintenance update

dlmiles:
GHA obsoleted macos-12 so we move to macos-13 for Intel.  Building with oldest Intel image to provide feedback of continued expected support on that platform, while ARM64 builds with latest to check the other end of their SDKs.

Lock the TCL version via brew to 8.6+ (currently the project is not expected to build on TCL9 until other work is completed) so this is causing build failure.

Added additional diagnostics to show the location of executables/DSOs if they got built from the tree.

Introduced GHA timeout clamping (MacOS default shell seems not to behave under CI as well as Linux does, re signal/pipe handling)

'brew install tcl-tk@8' appears to append the @8 to the installed path such as /usr/local/opt/tcl-tk@8 which configure needs to know.  configure_mac updated to detect this now TCL9 is the default version that maybe installed at the old location.  Then there is the issue of both TCL versions being installed side-by-side on the same system causing more issues over selection.

### DIFF
--- a/.github/workflows/main-macos.yml
+++ b/.github/workflows/main-macos.yml
@@ -10,8 +10,9 @@ on:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
-  simple_build_macos12:
-    runs-on: macos-12
+  simple_build_macos13:
+    runs-on: macos-13
+    timeout-minutes: 45 # x86_64 seems non-SSD based (slower)
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


RTimothyEdwards (approved):
This is all Mac-related stuff which all looks okay to me.  Presumably this needs revisiting for Tcl 9 support, but I will go ahead and merge this.

RTimothyEdwards:

PR #358, opened 2025-01-03
post K&R data to rodata segment

dlmiles:
Project directories that have received K&R to ANSI conversion have another tier of commits to tidy up globals and some data structures to move data from .data to .rodata segments and provide APIs using const keywords where possible.

### DIFF
--- a/calma/CalmaRdcl.c
+++ b/calma/CalmaRdcl.c
@@ -18,7 +18,7 @@
  */
 
 #ifndef lint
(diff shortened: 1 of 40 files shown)


RTimothyEdwards (approved):
All of this looks good.  All of the `rcsid[]` lines should just be removed. . .  The last update was in 2010, and eventually everything got moved to git.

RTimothyEdwards:
Pulled and merged to the opencircuitdesign.com repository.

PR #359, opened 2025-01-04
Issue 316 rework, Fix infinite def  write

dlmiles:
#316 reworked commit


RTimothyEdwards (approved):
This makes more sense to me and thanks to you (@dlmiles ) and @kareefardi for finding the root cause and debugging the issue.

RTimothyEdwards:
Pulled and merged to the opencircuitdesign.com repository.

PR #360, opened 2025-01-06
Various -Wall cleanups now compiler has better visibility to report more things

dlmiles:
Some are partial K&R removal as cleanup would have naturally happened.

Sometimes the cast has a space after and sometimes not, due to trying to fit style inline with the particular file.

Note the use of `cb_heap_kill_t` type, it is an intention to use/mark all callbacks like this with an official prototype.
This will also mark both all function implementation of the callback and the point it is invoked to allow an IDE to grep for this unique function pointer type, which is either used in code or in comments near code.


RTimothyEdwards (approved):
Looks good to me.  Noted the "FIXME" comment by "pid_t".

RTimothyEdwards:

PR #361, opened 2025-01-06
Setup as new PR (for resolution of #341)

dlmiles:
See #341 and #340


RTimothyEdwards:

PR #362, opened 2025-01-07
Make uses of 'bool' type compatible with C99 stdbool.h

dlmiles:
Can test for compiler feedback with -Wall and a minor modification:

```
# utils/magic.h
-typedef unsigned char bool;
+#include <stdbool.h>
+//typedef unsigned char bool;
```

Maybe there is a better name than RESTRUE ?  it just needed a name, this could probably have been defined as `(int)TRUE`

Not tested, so left in draft.


RTimothyEdwards (approved):
I'd have to stare at the code a while to figure out if there's a better substitute for "RESTRUE", so it will remain for now.

RTimothyEdwards:

PR #364, opened 2025-01-10
gcc: support -Werror=format-security

mtasaka:
Support gcc -Werror=format-security which is used by default on Fedora.

ext2spice/ext2spice.c: format not a string literal and no format arguments
extflat/EFargs.c: likewise
windows/windCmdNR.c: likewise


RTimothyEdwards:
Pulled and merged into the repository at opencircuitdesign.com.

PR #370, opened 2025-02-18
use-after-free/double-free concern CIFFreePath() non-invalidation

dlmiles:
The CodeQL analyser call graph of the theoretical use-after-free/double-free (due to non-invalidation) looks like:

```
1       pointer to free output argument         CIFrdutils.c:1343:12
2       path [Return]           CIFrdutils.c:1339:14
3       pointer to CIFFreePath output argument          CalmaRdpt.c:1160:18
4       *pathheadpp [Return]            CalmaRdpt.c:1113:15
5       calmaReadPath output argument           CalmaRdpt.c:240:24
6       pathheadp               CalmaRdpt.c:288:17
7       path            CIFrdutils.c:1339:14
8       path            CIFrdutils.c:1344:9
```

This is indicating the potential for `CIFFreePath()` will be called twice for the same memory block via this call graph (line numbers related to code pre-patch, might need to fuzz a little)

The other patch is a minor re-order (as this shows up now as a use-after-free when the freeMagic() delay-by-one-dealloc feature is disabled)

### DIFF
--- a/resis/ResMain.c
+++ b/resis/ResMain.c
@@ -1087,11 +1087,11 @@ ResExtractNet(node, goodies, cellname)
 		    resMakeDevFunc, (ClientData)thisDev);
 	if (result == 0)
 	{
-	    freeMagic(thisDev);
 	    TxError("No device of type %s found at location %d,%d\n",
 		    DBTypeLongNameTbl[thisDev->type],
 		    tptr->thisDev->location.p_x,
 		    tptr->thisDev->location.p_y);
+	    freeMagic(thisDev);
 	    continue;
 	}
 	thisDev->nextDev = DevTiles;


dlmiles:
Note this hasn't been tested (that is set to occur, with a bunch of other stuff over the next few weeks).

It is here (up on PR) to allow immediate visibility as maybe a domain knowledge expert can confirm from the review process is the solution is correct.

Maybe I should start to tag the status/quality of PRs on GH to communicate where along the process of quality/testing-status/impact-concern axis I place a particular PR to be.  Sometime publishing/visibility is more useful than the assumption of trying to get a PR merge ASAP.

RTimothyEdwards:
I don't think the solution is _wrong_, but I don't see how the existing code can result in a use-after-free.

As for the line reordering, fine, but I would think that there are about ten thousand places in the code that depend on the one-off deallocation, which is why I've never tried to mess with it myself, as much as I don't like it.

dlmiles:
The report:
```
1       pointer to free output argument         CIFrdutils.c:1343:12
2       path [Return]           CIFrdutils.c:1339:14
3       pointer to CIFFreePath output argument          CalmaRdpt.c:1160:18
4       *pathheadpp [Return]            CalmaRdpt.c:1113:15
5       calmaReadPath output argument           CalmaRdpt.c:240:24
6       pathheadp               CalmaRdpt.c:288:17
7       path            CIFrdutils.c:1339:14
8       path            CIFrdutils.c:1344:9
```

https://github.com/RTimothyEdwards/magic/blob/8.3.520/cif/CIFrdutils.c#L1343  free() call inside `CIFFreePath()`

https://github.com/RTimothyEdwards/magic/blob/8.3.520/cif/CIFrdutils.c#L1339  the argument `path` of the entry point that is the function `CIFFreePath(CIFPath* path)`

https://github.com/RTimothyEdwards/magic/blob/8.3.520/calma/CalmaRdpt.c#L1160  the site of the call to `CIFFreePath()` from inside `calmaReadPath()` this is the area my commit invalidates the pointer upon return

https://github.com/RTimothyEdwards/magic/blob/8.3.520/calma/CalmaRdpt.c#L1113  the argument `pathheadpp` of the entry point that is the function `calmaReadPath(CIFPath **pathheadpp, int iscale)` 

https://github.com/RTimothyEdwards/magic/blob/8.3.520/calma/CalmaRdpt.c#L240 the site of the call to `calmaReadPath()` using an address-of local variable

At this point we envisage the `pathheadp` currently points to a non-NULL invalidated memory, after line 240

But YES there is a return check for FALSE to then return from this function.   But the next item on the CodeQL list is:

https://github.com/RTimothyEdwards/magic/blob/8.3.520/calma/CalmaRdpt.c#L288  another call is made to `CIFFreePath()` but I can not see any route to get there.

So this can be considered a false positive report from CodeQL ?

dlmiles:
Merge Status: merge-on-hold  pending #384 review
Quality: ready as-is 99a693c4 enough to merge
Risk: low-impact, no expected behaviour change
Level of Testing: kick the tires (not targeted at this change-set)

dlmiles:
Merge Status: ready to merge

1 commit dropped here, so this is ready and separate update will be made on #384

RTimothyEdwards:
Pulled and merged.

PR #371, opened 2025-02-18
Always pass TileTypeBitMask by-pointer (not by-value)

dlmiles:
I assume this was to ensure the caller's copy of the mask is never accidentally modified, in the days before `const` and before putting read-only data in `.text` or `.rodata` segments.

No attempt has been made to optimize the receiving function to remove the need to copy it locally into its stack frame, and use the pointer directly (a number of scenarios exist to optimize on this).  I used an idiom that should not change existing behaviour and should work every time.

Just needed a resolution to not pass the large 32-byte wide object by-value (this is technically due to AVX2 data alignment requirements) elicits a GCC warning, also that passing large amounts of data by-value that may not be utilized at all by the receiving function (because a boolean decision skips it for this call) can not optimize the removal of the copy that takes place).

### DIFF
--- a/extflat/EFvisit.c
+++ b/extflat/EFvisit.c
@@ -33,7 +33,8 @@ static char rcsid[] __attribute__ ((unused)) = "$Header: /usr/cvsroot/magic-8.0/
 #include "utils/malloc.h"
 #include "utils/utils.h"
 #include "extflat/extflat.h"
-#include "extflat/EFint.h"
+#include "database/database.h"  /* before EFint.h for ArrayInfo */
+#include "extflat/EFint.h"      /* also has ArrayInfo */
 #include "tiles/tile.h"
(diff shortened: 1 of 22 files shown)


RTimothyEdwards:
I am okay with this change.

Note that the majority of the code is in sections like "plow" and "mzroute" and "iroute", which have been barely touched by me over the years because they're of questionable use.  There are definitely poor coding practices that have been cleaned up in other places but still exist in these lesser-known, lesser-used sections.

dlmiles:
Merge Status: ready to merge
Quality: complete no further work expected
Risk: low-impact, no expected behaviour change (just the header file include change expected to be highest risk part)
Level of Testing: non due to location of changes in project

RTimothyEdwards:
Pulled and merged.

PR #372, opened 2025-02-18
SUPPORT_DIRECT_MALLOC and SUPPORT_REMOVE_MALLOC_LEGACY

dlmiles:
The idiom created is like so to allow potential instrumentation/debugging possibilities around specific sites if it is found specific memory bugs need chasing, maybe due to the scoping choice of this patch (the location of the init() and end() sites) is found to be incorrect.
It also marks with a unique grep'able term 'freeMagic1' to facilitate later cleanup/removal of the idiom, or inspection for optimization at a later date.
The '1' in the naming related to the deferred-by-one more than it being version 1.
There is some overlapping usage so the variable naming ended up as `mm1` and `mm1_` where a 2nd variable was needed.

While this patch seeks to not modify magic default configuration (you must opt-in via manual editing of defs.mak after ./configure) it is designed to allow community testing, bug hunting and reporting to take place.

PLEASE DO NOT ENABLE THESE FEATURE FLAGS FOR PRODUCTION USE, only evaluation and bug hunting.

TODO: The tcl `version` command (and the startup banner) needs indicate which feature flags are enable for the specific binary being run.

--

Methodology summary:
Changes applied in commit order (except for last commit).
Use CodeQL to find use-after-free sites.
Introduce freeMagic1 idiom at each callsite, with an assessment of scope (placement of init/end).
Create the last commit in the sequence based on the changes.

---

Comment taken from the lead commit:

    This supports three build modes:
    
    No additional -D options, default legacy mode
    
    -DSUPPORT_DIRECT_MALLOC, magic will use direct calls to libc malloc/free
     and will leave in place the symbols now renamed as mallocMagicLegacy()
     freeMagicLegacy() and callocMagicLegacy().
    
    -DSUPPORT_DIRECT_MALLOC -DSUPPORT_REMOVE_MALLOC_LEGACY as above but will
     remove the three legacy functions from the binary to provide assurance
     they can not be used.
    
    The system malloc is thread-safe the legacy magic malloc has a global
    deferred free pointer and the mmap() allocate has a free-list that is
    not thread-safe making use of free not thread-safe.
    This could of course be improved with the use of
    atomic_compare_and_exchange operations but for what gain ?
    
    Then there is the additional function call overhead (of the indirection)
    and a few tests/branches inserted into a commonly used code paths around
    memory allocation, it hides the call site of the malloc/free usage from
    the compiler which maybe have special optimization cases.
    
    The existing malloc/free makes static code analysis around memory
    allocation more problematic, also use of runtime analysers will operate
    better with a fail-fast to bad memory usage.

### DIFF
--- a/bplane/bpBins.c
+++ b/bplane/bpBins.c
@@ -93,7 +93,7 @@ static BinArray *bpBinArrayNew(int dx,      /* x diameter of bins */
 
   /* allocate array */
   size = sizeof(BinArray) + numBins*(sizeof(void *));
-  new = (BinArray *)callocMagic(size);
+  new = (BinArray *)callocMagic(1, size);
 
   /* initial */
   new->ba_bbox = *bbox;
--- a/calma/CalmaRdpt.c
+++ b/calma/CalmaRdpt.c
@@ -360,12 +360,14 @@ calmaElementBoundary(void)
     }
 
     /* Paint the rectangles (if any) */
+    free_magic1_t mm1 = freeMagic1_init();
     for (; rp != NULL ; rp = rp->r_next)
     {
 	if (plane)
 	    DBPaintPlane(plane, &rp->r_r, CIFPaintTable, (PaintUndoInfo *)NULL);
-	freeMagic((char *) rp);
+	freeMagic1(&mm1, (char *) rp);
     }
+    freeMagic1_end(&mm1);
 
     if (cifCurReadPlanes == cifEditCellPlanes)
     {
--- a/calma/CalmaWrite.c
+++ b/calma/CalmaWrite.c
@@ -1400,8 +1400,10 @@ calmaOutFunc(
 	    {
 		pllist[i].pl_label = ll->ll_label;
 		pllist[i].pl_port = (unsigned int)ll->ll_attr;
-		freeMagic(ll);
+		free_magic1_t mm1 = freeMagic1_init();
+		freeMagic1(&mm1, ll);
 		ll = ll->ll_next;
+		freeMagic1_end(&mm1);
 		i++;
 	    }	
 
@@ -2422,19 +2424,26 @@ calmaProcessBoundary(
 
 	/* Free the LinkedBoundary list */
 
-	lbref = listtop;
-	while (lbref->lb_next != listtop)
 	{
-	    freeMagic(lbref);
-	    lbref = lbref->lb_next;
+	    free_magic1_t mm1 = freeMagic1_init();
+	    lbref = listtop;
+	    while (lbref->lb_next != listtop)
+	    {
+		freeMagic1(&mm1, lbref);
+		lbref = lbref->lb_next;
+	    }
+	    freeMagic1_end(&mm1);
 	}
-	freeMagic(lbref);
     }
 
     /* Free the BoundaryTop list */
 
-    for (bounds = blist; bounds != NULL; bounds = bounds->bt_next)
-	freeMagic(bounds);
+    {
+	free_magic1_t mm1 = freeMagic1_init();
+	for (bounds = blist; bounds != NULL; bounds = bounds->bt_next)
+	    freeMagic1(&mm1, bounds);
+	freeMagic1_end(&mm1);
+    }
 }
 
 /*
@@ -2500,8 +2509,10 @@ calmaMergePaintFunc(
 		lb = edge;
 		while (lb->lb_next != edge) lb = lb->lb_next;
 		lb->lb_next = edge->lb_next;
-	 	freeMagic(edge);
+		free_magic1_t mm1 = freeMagic1_init();
+		freeMagic1(&mm1, edge);
 		edge = edge->lb_next;
+		freeMagic1_end(&mm1);
 	    }
 	}
 	else
@@ -2720,11 +2731,13 @@ done_searches:
 
 	    if (num_points != 4)
 	    {
+		free_magic1_t mm1 = freeMagic1_init();
 		for (i = 0; i < num_points; i++)
 		{
-		    freeMagic(edge);
+		    freeMagic1(&mm1, edge);
 		    edge = edge->lb_next;
 		}
+		freeMagic1_end(&mm1);
 		edge = NULL;
 	    }
 	    if (!StackEmpty(SegStack))
--- a/calma/CalmaWriteZ.c
+++ b/calma/CalmaWriteZ.c
@@ -1324,8 +1324,10 @@ calmaOutFuncZ(
 	    {
 		pllist[i].pl_label = ll->ll_label;
 		pllist[i].pl_port = (unsigned int)ll->ll_attr;
-		freeMagic(ll);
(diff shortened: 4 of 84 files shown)


RTimothyEdwards:
As I mentioned in another PR comment, I don't like the one-off deallocation but since it exists in ten thousand places in the code, I have just learned to live with it instead of trying to get rid of it.  If this can be proven to catch all cases and not break anything anywhere, I'm perfectly fine to be done with the one-off deallocation, which has been, by and large, a royal pain in the neck.

dlmiles:
I'll keep you posted, its too early to say.  It is possible to apply this PR (build so libc malloc/free is used directly), then fire up magic under valgrind and do some things and then shutdown without any reported issues.
So if that is possible things can not be that bad. :)

It looks like the static code analysis worked well enough to immediately spot all the locations I was looking for first time.

I will take any pointers to ZIP projects and scripts to run to exercise things more thoroughly, I can run under valgrind/address-sanitizer/gdb over several days to find issues (or gain confidence).
Would definitely appreciate some kind of ZIP+scripts that demonstrate and exercise some of the lesser known features.
Maybe that would be good form of documentation for those features?
I am trying to create a project-of-projects that attempt code coverage.

RTimothyEdwards:
If that were easy I would have created a comprehensive CI by now.
The feature set of magic is vast and it is a lot of work to exercise all the parts of the code, which covers everything from DRC and extraction to GDS generation to plotting and routing and interactive wiring. . . 
I saw recently a video of a circuit being drawn that was related to Tiny Tapeout and likely (?) came from somebody using the (recently fixed) `logcommands` feature to save the entire layout process.  Getting hold of such a command set would make a good coverage of interactive commands.  For non-interactive stuff, running CACE on any of the sky130 analog blocks that support it (which is now most or all of them) would capture most of the features like DRC, extraction, and GDS generation, although (as I've pointed out before) some of the features like "plow" and "iroute" are so old that they are rarely used by anyone and hard to find examples for.  The original magic tutorial set might be the best place to go to exercise those features.

RTimothyEdwards:
Uri Shaked might know something about the magic layout video for TinyTapeout.

dlmiles:
Merge Status: could be merged, but not expected to be merged, review this status in 2 months
Quality: usable as is (for example if memory bug chasing, this might provide a different angle and fail-fast to find the source of the bug easier)
Risk: very-high-impact, this may patch the 1st-order delayed-free use cases and appear to work, but I expect it will now uncover 2nd-order issues we don't know about yet
Level of Testing: valgrind with basic usage (extensive testing will take place over the coming months)
TODO: startup banner, `--version`, `version` command, should clearly indicate the feature flags a build was created with to help support issue around its use

It is expected that bugs found from it will be the subject of new/separate PRs, but it may depend on the bug.

dlmiles:
1cad1527 updated with 1st order issue CodeQL static analyser maybe not have spotted.

dlmiles:
@RTimothyEdwards Looking to remove the draft status on this, opening this for release.  Just need feedback this doesn't break the world for MacOS.

Recap:
 * it seems the original commit was correct (after the `efConnectionFreeLinkedList()` mopup change)
 * this does not change the default build (you have to opt-in manually after configure and read the comments in `defs.mak` to engage the malloc changes)
  * the MacOS in CI shows startup failures (for some commands), which relate to `Kick the Tyres` testing not the building phase:
   `dyld[25928]: missing symbol called` is the unusual item seen in CI output, it is not clear what is missing

@d-m-bailey
Any Mac users with time to test compile, install and use to confirm this PR doesn't break normal usage.

```
# example of what the test compile setup looks like
cd /tmp
git clone https://github.com/RTimothyEdwards/magic
cd magic
git describe --long

git fetch origin pull/372/head:refs/heads/pull/372
git checkout pull/372
git describe

# follow your normal process here, maybe it looks like:
./scripts/configure_mac
make
make install

# the commands in CI showing error are the unusual sequence:
echo "=== magic -d null -noconsole -nowindow -T scmos"
echo "version ; quit" | magic -d null -noconsole -nowindow -T scmos
echo "=== magic -d null -noconsole -T scmos"
echo "version ; quit" | magic -d null -noconsole -T scmos
```

Then run it normally and see if it crashes on startup or exit.
The error `dyld[25928]: missing symbol called` seen in GitHub CI doesn't show any information to see the symbol, but this is expected on startup as it bails.
Then extended test to see if any MacOS specific issues under normal use.

dlmiles:
The code as-is does not remove the `mallocMagic()` it still requires manual editing (by uncommenting of `defs.mak` after `./configure`).

I am confident this caught all the parts relating to delay-by-one expectation, but the patch is constructed to allow 4 phases to the lifecycle of this change:
* Experimental (as now/previously), that is you need to apply patch from github to use it.
* Available via edit of `defs.mak` after `./configure`.  This allows valgrind and other patches on top to be merged.  This may stay like this for some months, maybe 6 months.  This phase is designed to allow things to be tried easily.
* Available by default to users, which is just a patch to edit the `defs.mak.in` to enable by default.  This may stay like this for 1 year, allowing anyone with an issue to easily revert and report that something works when reverted, but not by default configure.  This phase is designed to allow users to report issues easily via a recompile.
* Cleanup, which is the removal of the legacy `mallocMagic()` and a global search/replace of all call sites.  It would be expected via a new patch to remove the legacy code and modify all the call sites and cleanup the project via a different patch set than this PR.  This patch does not exist at this time as it is so far out into the future.

The guideline is to use malloc/free/calloc like normal C language usage, however this can be done during/at phase 4 there is no hurry to adopt today, however you'd expect new code to not be created to rely on the delay-by-one behaviour and manage that itself, so if the code does not need delay-by-one behaviour you can use system allocators directly:
* `mallocMagic(1234) => malloc(1234)`
* `freeMagic(ptr) => free(ptr)`
* `callocMagic(1234 * 4) => calloc(4, 1234)`  the limited uses have had their API modified to follow libc standard, that is contained in this PR.

Until we are at phase4 you can continue to use mallocMagic() etc..  The purpose of advancing to phase2 is to allow the work on top to be able to be merged.

One issue with the old mallocMagic() is that it is not thread-safe, you can't use the legacy APIs from different threads without expecting the defer-by-one pointer to be lost/corrupted.  It is a shared global, it could be fixed via use of atomic 64bit, but there seems little point.  At the moment limit all mallocMagic/freeMagic to the main thread.

dlmiles:
github shows it as no conflicts, from a `will it merge ok?` point of view.

it does not matter so much if there are further uses of `mallocMagic/freeMagic` added (since original patch), only the use of `freeMagic` where a deferred-by-one is a requirement would add the pattern to manage the lifetime requirement locally by the thread (on the stack).   I don't think this is the case. But I check over that when there there are two commit-id's known to check between.   Easy enough to grep for inserted `freeMagic()` since the original time of the patch.

RTimothyEdwards (approved):
Okay, looks good and works on my end (merge was clean, although I have not tried enabling the modified behavior, just wanted to make sure that existing behavior isn't broken).

PR #380, opened 2025-02-18
DBtiles.c: DBFreePaintPlane() remove the TiFree() deferred assumption

dlmiles:


dlmiles:
The intention now is to add valgrind specific hints via `#include <valgrind/memcheck.h>` into the TiAlloc() custom allocator, to the memory checker can good the heavy lifting in validating (for finding the next problem).  Also poisoning the data on TIFree().

Not sure if you want that (valgrind hints) as an additional commit with ifdef around ?

There is already `K&R tiles/**` ready here (just not on PR), so I may push that before releasing this.


Merge Status: merge on hold (pending valgrind style testing, review status in 2 months)
Quality: ready to merge
Risk: low-impact (this is rarely used deallocation function area)
Level of Testing: none (due to expecting to valgrind it)

RTimothyEdwards:
Yes, I would like to have an ifdef around this, and a check for valgrind and the memcheck.h file in the configure script.

dlmiles:
Sorry my text was not more clear.  Yet any commit adding valgrind support will ifdef around the parts (obviously not every build wants to include it).  The focus of my question was not in use of ifdef or not-use-ifdef.

The question is more should such a commit be in the main project at all.

I am currently using with a manual edit of `defs.mak` just like SUPPORT_DIRECT_MALLOC requires.  I make the patch so all you have to do is run configure and then manually edit defs.mak to uncomment an existing line.  i.e. there is no automatic way to enable it (even by accident).

If autoconf was used for valgrind maybe it would be a `--enable-valgrind` which will then perform prereq checks and also enable it.

-- 

Merge Status: this change-set is ready now
Quality: ready to merge
Risk: low-impact (this is rarely used deallocation function area)
Level of Testing: valgrind testing confirms this change-set as-is works ok and an example issue found can be seen at #394

Note this change-set does not add valgrind support (as a result of the dialog here that will be the subject of a separate change-set)

dlmiles:
This is how this is looking now to maintain code clarity at the point of use, while also not impinging on future multiple-threaded modifications to the same database, or multiple-thread writing to different databases inside the same process.

```
extern void TiFreeIf(Tile *tile);
extern void TiFree1(Tile **delay1, Tile *tile);
```

Attempts to use inline (so the compiler can optimize at the point of use).

Re my comments on valgrind support will be the subject of a different PR.

dlmiles:
> @dlmiles :  Is this pull request still relevant or has it been superceded?
> It looks like PR #414 contains all of the changes in #380 so I will just merge 414.

Yes that is correct #414 should include these commits. As the groundwork here is needed by #414.

RTimothyEdwards:
Closing this PR prior to merging #414.

PR #383, opened 2025-02-18
K&R lef/**

dlmiles:


dlmiles:
Update is correction to accidental comment edit from semi-automatic K&R conversion.

Then some matters arising I could not resolve at the time, with LefNextToken() constification, which needed a but more attention to understand what change is needed,  and then finally I see what was left in .data to move to .rodata cleanup.

dlmiles:
Only functionality change is introduction of LefHelper_DBTechNameType_LefLower() which technically changes that the original `token` is not lower cased after the point of use, but this did not look to impact anything I could see.
Also use of `tolower()` in a place to detect 'horizontal`
Both these in the LefNextToken() commit near the most recent.

Merge Status: ready to merge
Quality: ready to merge
Risk: low-impact (should not change much functionality - see above)
Level of Testing: kick the tires (high level non-targeted)

dlmiles:
840e58b6f11c4dc6530987c722cc7dd3d3e36cfd  ASSERT(cond) was inverted fix

RTimothyEdwards (approved):
Looks good to me.  Feel free to go through the "FIXME" lines and remove the lines that are not being used.  This is just my failure to clean up properly after writing and editing large amounts of code.

RTimothyEdwards:
Pulled and merged (as noted, I have not done anything about the "FIXME" lines, but I think they should all be deleted as suggested).

PR #384, opened 2025-02-19
RFC re alternative approach to #370

dlmiles:
This PR is a Request For Comments on these alternative commits related to #370 

In PR #370 I invalidate with `*pathheadpp = NULL` but this is of course extra work the CPU has to do.

So here is a V1 and V2 alternative, that achieves multiple things:

* make both functions easier for the next person to understand/reason about (understand this maybe somewhat opinionated)
* reduces the amount of work the CPU needs to do overall (true of V2 not of V1), mainly from the management on both sides of the function call for having 2 piece of data, return value and CIFPath*
* -O3 compiler headline in respective commit comment (of the function implementation in situation inside magic)
* quashes (false positive) static code analysis report concerning this (this is an issue, in that its easier to do something in the code than manage the extensive lists of points picked up)

just making a case to understand what kind of factors justify

At the moment consider the patch untested, looking to see if there is understanding that in win-win-win (readability-performance-concern_management) scenarios.
If there is a positive feedback I shall get to testing and report back in issue (could be days/weeks)
Thanks

### DIFF
--- a/calma/CalmaRdpt.c
+++ b/calma/CalmaRdpt.c
@@ -58,7 +58,7 @@ extern int CalmaPathCount;
 extern HashTable calmaDefInitHash;
 
 extern void calmaLayerError(char *mesg, int layer, int dt);
-bool calmaReadPath(CIFPath **pathheadpp, int iscale);
+CIFPath *calmaReadPath(int iscale);
 
 /*
  * ----------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


dlmiles:
PR currently exists for feedback would welcome review, to understand alternative viewpoint around change of this nature.

Merge Status: merge on hold (discussion only)
Quality: good enough to dicuss
Risk: low-impact (should not change any functionality)
Level of Testing: will receive targeted testing pending a favourable review to proceed

A favourable review here would have me amend #370  to remove `*pathheadpp = NULL` changes and push the reorder on its own.  Then rework this PR to be production ready after testing.
An unfavorable review here would have me close this and remove the `*pathheadpp = NULL` changes and push the reorder on its own.

RTimothyEdwards (approved):
After reviewing the code, I think this approach is fine.  I don't see any way in which the result is not redundant to the check of pathheadp being NULL or non-NULL.

RTimothyEdwards:
Pulled and merged.

Issue #381, opened 2025-02-18
ext2spice.c mergeAttr() request for review


dlmiles:
https://github.com/RTimothyEdwards/magic/blob/c4a2a54cb77aea8ce8f09c23f5afd12b8ac6ec99/ext2spice/ext2spice.c#L4259

``
void
mergeAttr(a1, a2)
    char **a1, **a2;
{
    if (*a1 == NULL)
	*a1 = *a2;
    else
    {
	char *t;
	int l1 = strlen(*a1);
	int l2 = strlen(*a2);
	t = (char *) mallocMagic((unsigned int)((l1 + l2) + 1));
	t = (char *) strcat(*a1, *a2);  // PLEASE REVIEW THIS AREA
	freeMagic(*a1);
	*a1 = t;
    }
}
``

Consider:

``
	char *t;
	int l1 = strlen(*a1);
	int l2 = strlen(*a2);
	t = (char *) mallocMagic((unsigned int)((l1 + l2) + 1));
#if 0
 	t = (char *) strcat(*a1, *a2);
#else
	strcpy(t, *a1);
	strcat(t, *a2);
#endif
	freeMagic(*a1);
	*a1 = t;
``

RTimothyEdwards:
I agree the original version doesn't make any sense;  the `strcat()` function merges string `a2` into string `a1` which does not necessarily have room to hold it, and the returns a pointer to string `a1`, causing the allocated memory `t` to be orphaned.  I suggest fixing this as you have done above.

Likewise, I can find no place that calls `mergeAttr()` in `ext2sim` so I agree with your assessment of just removing it altogether.

PR #390, opened 2025-02-24
mergeAttr() fix and unused macro removal

dlmiles:
See #381

### DIFF
--- a/ext2sim/ext2sim.c
+++ b/ext2sim/ext2sim.c
@@ -1744,34 +1744,6 @@ Dev    *dev;
 /*
  * ----------------------------------------------------------------------------
  *
- * mergeAttr --
- * Macro to merge two attribute strings
- *
- * Results:
- *  The merged strings
- *
- * Side effects:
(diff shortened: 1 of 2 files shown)


RTimothyEdwards (approved):
Looks good to me.  It might have implications if ext2spice and ext2sim are compiled as independent targets, but I have not done that for ages and it may not be possible any more anyway, and is not worth supporting anyway.

RTimothyEdwards:
Pulled and merged.

dlmiles:
> Looks good to me. It might have implications if ext2spice and ext2sim are compiled as independent targets, but I have not done that for ages and it may not be possible any more anyway, and is not worth supporting anyway.

I have a Makefile/build overhaul branch around, I think part of that was to fix those targets building by default.  The idea is that everything that is possible to be built gets built by default, so CI works better to detect commits that break building everything.  Just so such issues are visible sooner and not only when a user reports and draws attention to the scenario.

PR #391, opened 2025-02-24
RLIMIT_NOFILE >= 4096 when on TCL9 support

dlmiles:
Including defensive counter measures for areas of magic that don't support FDs >= 1024.

Any suggestions on getting to see TxGetInputEvent() in action ?  Does not appear to be in use ?

Comments around this item https://github.com/RTimothyEdwards/magic/pull/374#issuecomment-2668749478

### DIFF
--- a/graphics/grMain.c
+++ b/graphics/grMain.c
@@ -457,8 +457,15 @@ grFgets(str, n, stream, name)
     twentySecs.tv_sec = 20;
     twentySecs.tv_usec = 0;
 
+    const int fd = fileno(stream);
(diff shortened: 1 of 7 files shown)


dlmiles:
minor change to fix text in printf and make tclmagic.c conform to file lint style

dlmiles:
rebased onto current master, pushed 1 more commit

dlmiles:
a742413 adjustment to ifdef placement to account for platforms without rlimit so the type `rlim_t` will not be available

RTimothyEdwards (approved):
Looks good to me, as far as I understand what's being done.

RTimothyEdwards:
Pulled and merged.

PR #392, opened 2025-02-24
K&R sim/**

dlmiles:


RTimothyEdwards (approved):
Looks good, although I don't really understand what's going on with the comments about "cb_database_srpaintarea_t";  these look like compiler hints?

RTimothyEdwards:
Pulled and merged

dlmiles:
> Looks good, although I don't really understand what's going on with the comments about "cb_database_srpaintarea_t"; these look like compiler hints?

The type: `typedef (*cb_database_srpaintarea_t)(Tile *tile, ClientData cdata)`  is the official type for the callback method for `DbSrPaintArea()`

When converting everything to ClientData additional comment information will be added to retain the actual expect type of the `ClientData` at any location where the information would otherwise be lost due to the conversion.  The strict use of the prototype for the purpose of symbol linkage was possible before in K&R because there was no argument prototype available to check.

When completed (the current various domains of changes) The invocation point (of the callback) is marked `/** @invoke cb_label_t`, each callback function is marked `/** @typedef cb_label_t */`, each forward declaration is marked.  Such that it is possible to grep the codebase for all uses of a particular callback.

The use of comments in the form of `/** @something some information */` is more an automated documentation convention.  At this time they are directed at the human reader, to indicate and help locate (via grep) all the callbacks used by a particular invocation.  However the form can also be used by other tooling like static-code-analysis (since the double asterisk comment and at-sign prefixed tag) is somewhat of a convention across multiple C/C++ like languages that use this comment form.  So there is usually plugin assistance available to detect presence and parse.

My tree locally may have branches with other conversions that have already occurred and some of the changes to this change-set needed some of that other work to be present to close off and complete the subdir `sim/**` which is why they appeared in this change-set.  Because for example the function signature was changed of to add `const` and the only way to have the compiler validate the entire project is still correct is to convert the affected functions from K&R into ANSI so the function prototype is complete and visible.  So this is why some random function in database module were converted, when in my tree locally there is a branch with most of `database/**` already converted.

PR #393, opened 2025-02-26
ExtTech.c: fix uninitialized exts_linearResist value leaks into extresist

dlmiles:
This value appears to be initialised at only one spot in the codebase (under very narrow conditions) but extresist will read it and make branching decisions based on the uninitialised state.

This 'X' state propagation appears to eventually get processed in ResWriteExtFile() near where final output formatting is occurring.

It is unclear (at this time) if it perturbs output values in a problematic way, or if due to algorithmic reasons the data is discarded before output anyway.  I have at least one trace run (with multiple triggers) of printf formatters handling uninitialised data in ResWriteExtFile().

### DIFF
--- a/extract/ExtTech.c
+++ b/extract/ExtTech.c
@@ -2807,6 +2807,7 @@ ExtTechLine(sectionName, argc, argv)
 		    }
 		    devptr->exts_deviceResist.ht_table = (HashEntry **) NULL;
 		    HashInit(&devptr->exts_deviceResist, 8, HT_STRINGKEYS);
+		    devptr->exts_linearResist = 0;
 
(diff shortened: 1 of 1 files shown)


dlmiles:
Just to update on this after running more analysis:

https://github.com/RTimothyEdwards/magic/blob/705b4da105824102a31f19df6bb6b9b583555567/resis/ResRex.c#L1851-L1852

It is this totalization summary that is perturbed by the data in the few project samples I have tried so far.  This particular line is only printed to the logs.  That doesn't rule out the main output being affected, but I have no example on that to show.

RTimothyEdwards (approved):
I'll assume this is the correct solution.

RTimothyEdwards:
Pulled and merged.

PR #405, opened 2025-06-12
issue#404 TiJoin[XY] use-of-free mitigation

dlmiles:


RTimothyEdwards (approved):
That does appear to fix the problem, thanks!

I am going to just make the change manually and not as a pull request, because I want to give each routine a header in the standard style of the code.

RTimothyEdwards:
This has been merged into the code manually and will be mirrored to github overnight.  Thank you for solving this!

PR #407, opened 2025-06-27
AppImages for all occasions

dlmiles:
This fixes the recent commit that breaks older GCC (AppImage EL7)
Upgrades Tcl 8.6 to latest (EL7/EL8)
This adds EL8/EL9/EL10 builders.
EL9/EL10 moves to Tcl9
Improve release notes/info.


dlmiles:
Example can be seen here https://github.com/dlmiles/magic/releases until I delete/cleanup.

RTimothyEdwards (approved):
Dealing with workflows and appimages is not my specialty, so I will assume it is correct, and from a cursory glance it all looks reasonable.  The cleanup of code in tiles/tile.h probably belongs in a different commit, but I have no issue with it.

RTimothyEdwards:
Pulled and merged.

dlmiles:
> Dealing with workflows and appimages is not my specialty, so I will assume it is correct, and from a cursory glance it all looks reasonable. The cleanup of code in tiles/tile.h probably belongs in a different commit, but I have no issue with it.

The tile.h is to fix the appimage7 builder (that has been broken for a few months, ~5 releases). 

Because GCC on CentOS7 is older and does not support the style of `inline` in the removed lines, but newer compilers do support it.

PR #412, opened 2025-07-12
DBcellsrch.c: #ifdef ROUTE_MODULE around hook to update MZAttachHintPlanes()

dlmiles:
DBcellsrch.c: #ifdef ROUTE_MODULE around hook to update MZAttachHintPlanes()

When building without this module MZInit() does not get called and MZAttachHintPlanes() does not guard against that situation.

To allow unused code to be pruned from final binary, better to remove the cross module hook breaking the dependency.

### DIFF
--- a/bplane/bpDump.c
+++ b/bplane/bpDump.c
@@ -78,6 +78,7 @@ void bpDumpRect(Rect *r)
     fprintf(stderr,"%d",
 	    r->r_ytop);
   }
+#ifdef CIF_MODULE
   else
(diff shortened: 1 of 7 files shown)


dlmiles:
Found 3 addition cherry-pick on the same topic of missing ifdef

RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #414, opened 2025-07-15
issue#404 v2 This is the cleaned up version

dlmiles:
This is built on top of #380 (requires autoconf INLINE, and maybe tile.h TiFreeIf() introduction etc...)

This will build differently depending on your -O use as it makes use of inline and the compiler can decide when to optimize.

### DIFF
--- a/tiles/tile.c
+++ b/tiles/tile.c
@@ -23,6 +23,9 @@ static const char rcsid[] __attribute__ ((unused)) = "$Header: /usr/cvsroot/magi
 #include <stdio.h>
 #include <unistd.h>
 #include <stdlib.h>
+#ifdef HAVE_SYS_MMAN_H
+#include <sys/mman.h>
+#endif
 
 #include "utils/magic.h"
 #include "utils/malloc.h"
--- a/tiles/tile.c
+++ b/tiles/tile.c
@@ -853,6 +853,8 @@ TiFree(
 /* Use of 'extern inline' force an emit of inline code at a symbol */
 extern inline void TiFreeIf(Tile *tile);
 extern inline void TiFree1(Tile **delay1, Tile *tile);
+extern inline void TiJoinX1(Tile **delay1, Tile *tile1, Tile *tile2, Plane *plane);
+extern inline void TiJoinY1(Tile **delay1, Tile *tile1, Tile *tile2, Plane *plane);
 #else
 /* To support older compilers (that don't auto emit based on -O level) */
 void
@@ -867,6 +869,20 @@ TiFree1(Tile **delay1, Tile *tile)
 {
     TiFreeIf(*delay1);
     *delay1 = tile;
+}
+
+void
+TiJoinX1(Tile **delay1, Tile *tile1, Tile *tile2, Plane *plane)
+{
+    TiFreeIf(*delay1);
+    TiJoinX(tile1, tile2, plane);
+}
+
+void
+TiJoinY1(Tile **delay1, Tile *tile1, Tile *tile2, Plane *plane)
+{
+    TiFreeIf(*delay1);
+    TiJoinY(tile1, tile2, plane);
 }
 #endif
 
--- a/tiles/tile.c
+++ b/tiles/tile.c
@@ -61,7 +61,7 @@ global const Rect TiPlaneRect = { {MINFINITY+2, MINFINITY+2}, {INFINITY-2, INFIN
 
 #ifdef HAVE_SYS_MMAN_H
 
-static Tile *TileStoreFreeList = NULL;
+Tile *TileStoreFreeList = NULL;
 
 /* The new Tile Allocation scheme (Magic 8.0) */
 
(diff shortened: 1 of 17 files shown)


tiles/tile.c
```diff
@@ -868,6 +870,20 @@ TiFree1(Tile **delay1, Tile *tile)
     TiFreeIf(*delay1);
     *delay1 = tile;
 }
+
+void
+TiJoinX1(Tile **delay1, Tile *tile1, Tile *tile2, Plane *plane)
+{
+    TiFreeIf(*delay1);
+    TiJoinX(tile1, tile2, plane);
```
dlmiles, line 791: *delay1 = tile2; ???

tiles/tile.c
```diff
@@ -868,6 +870,20 @@ TiFree1(Tile **delay1, Tile *tile)
...
+}
+
+void
+TiJoinY1(Tile **delay1, Tile *tile1, Tile *tile2, Plane *plane)
+{
+    TiFreeIf(*delay1);
+    TiJoinY(tile1, tile2, plane);
```
dlmiles, line 798: *delay1 = tile2; ???

tiles/tile.h
```diff
@@ -279,10 +279,26 @@ TiFree1(Tile **delay1, Tile *tile)
     TiFreeIf(*delay1);
     *delay1 = tile;
 }
+
+inline void
+TiJoinX1(Tile **delay1, Tile *tile1, Tile *tile2, Plane *plane)
+{
+    TiFreeIf(*delay1);
+    TiJoinX(tile1, tile2, plane);
```
dlmiles, line 316: *delay1 = tile2; ???

dlmiles:
No wonder it will test ok, it is safe like Rust and just leaks.

RTimothyEdwards (commented):
Your questions about "*delay1 = tile2; ???"---Can you clarify if this is something you are unsure about, or something that you think needs to be added?  Should I approve and merge without this resolved?

dlmiles:
Something that needs to be added, but shall retest (with tile supported valgrind).

The current code looks like it will leak the Tile that should have been freed, because it is never freed.  So while this tested with tile supported valgrind (and also tested ok with your test case), I was looking for more serious use-after-free issues, not memory leaks (as tile deallocation on exit is not -DPURIFY perfect, trying to fix leaks is not straightforward at this time).

This patch as-is is ok (safe).  But the intent was to include `*delay1 = tile2;` in the 4 locations I indicated.  But that change will need to be valgrind tested (by me) before general use.

Leaks are a much lower priority issue, than use-after-free.

RTimothyEdwards (approved):
From comments by Darryl, I'm accepting this with the understanding that it is not perfect, but is probably still an improvement upon the existing code.

RTimothyEdwards:
Pulled and merged.

RTimothyEdwards:
@dlmiles :  Now magic just crashes whenever I load a layout.  Doesn't seem to happen on simple test layouts, but for example if I run magic on caravel.mag from github.com/fossi-foundation/caravel, then I get an immediate crash on loading.

RTimothyEdwards:
@dlmiles :  Continuing to test on cells from caravel:  Loading "caravel.mag" segfaults immediately.  Loading "chip_io.mag" does not segfault immediately but segfaults if I expand the subcells.  Loading "simple_por" is fine and I can expand the subcells and there is no issue (although I have not tried editing or extracting, for example).

RTimothyEdwards:
@dlmiles:  This is what I'm getting when loading "chip_io.mag" from caravel and expanding:
```
Thread 1 "wish" received signal SIGSEGV, Segmentation fault.
0x00007f2833c579eb in dbNMEnumFunc (
    tile=<error reading variable: Cannot access memory at address 0x7ffdc4461ff8>,
    arg=<error reading variable: Cannot access memory at address 0x7ffdc4461ff0>)
    at DBpaint.c:1804
1804    {
(gdb) up
#1  0x00007f2833c6784f in DBSrPaintArea (hintTile=0x7f2831756220, plane=0x55b10263a5b0,
    rect=0x7ffdc4462220, mask=0x7f2834743ac0 <DBAllTypeBits>,
    func=0x7f2833c579e3 <dbNMEnumFunc>, arg=0x7ffdc4462108) at DBtiles.c:421
421                         if ((*func)(tp, arg)) return (1);
(gdb) print tp
$1 = (Tile *) 0x7f2831763c70
(gdb) print func
$2 = (int (*)()) 0x7f2833c579e3 <dbNMEnumFunc>
(gdb) print arg
$3 = (ClientData) 0x7ffdc4462108
```
So the values of `tp` and `arg` are not being passed correctly to function `dbNMEnumFunc()`.

RTimothyEdwards:
@dlmiles :  Further investigation shows that it is apparently related to "stdbool".  On my system, it broke between these two commits:
```
commit 7029971c33c3c7b4e0227e57642653b64bd2f9a8
Author: Darryl L. Miles <darryl.miles@darrylmiles.org>
Date:   Wed Jun 18 14:31:17 2025 +0100

    configure: autoconf regen (2.69) AC_HEADER_STDBOOL

commit bfbdf45b88c9f5e94bbc4db031db86387f1de30d
Author: Darryl L. Miles <darryl.miles@darrylmiles.org>
Date:   Wed Jun 18 14:29:37 2025 +0100

    C23: use AC_HEADER_STDBOOL #include <stdbool.h>
```
Somewhere, a misunderstanding of the size of type "bool" is causing a misalignment of subroutine arguments.

RTimothyEdwards:
Sorry, did not mean to reopen this issue.  Closing it again and reopening the one that broke the app.

Issue #413, opened 2025-07-14
fail to build magic on fedora 42


jalcim:
```
[...]

PR #418, opened 2025-07-17
fix GrTextSizePtr API interface return type consistency

dlmiles:
commit 86b5d591d from 20241004 by me, modified return type of GrTextSize API in the graphics from 'void' to 'int' to convey and error scenario to indicate when 'r' was not filled in for the caller, but this is a multi graphics driver API interface so requires all downstream graphics engines to also support the return type change.

This API can error and indicates when 'Rect *r' was successfully updated.

See also #413  this error message relates to this return type inconsistency (which shows up with some build configurations that are also shown as failures in the CI-canary-matrix)

### DIFF
--- a/graphics/grMain.c
+++ b/graphics/grMain.c
@@ -158,7 +158,7 @@ void (*GrClosePtr)()		= NULL;
 void (*GrSetCMapPtr)()		= NULL;
 
 void (*GrSetCursorPtr)()	= NULL;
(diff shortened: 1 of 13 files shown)


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #424, opened 2025-07-24
cif_templates: Fix Makefile race condition with objs directory creation

tskaar:
Hi!

This fixes some of the racy-ness I have seen when building on Arch Linux, where one ends up not packaging/copying over the techfiles from SCMOS as an effect. 

Based on the build-logs the reason was that the objs_dir wasn't created when starting to create the targets, hence forcing objs_dir as a dependency for each target should resolve this racy-ness

### DIFF
--- a/scmos/cif_template/Makefile
+++ b/scmos/cif_template/Makefile
@@ -31,36 +31,36 @@ scg: cifout.c cifout-cmos26b.gen cifout.gen cifout.nw\
 $(OBJS_DIR):
 	$(MKDIR) $(OBJS_DIR)
 
-$(CIFIN): cifin.c cifin-cmos26b.gen cifin.gen cifin.nw cifin.oldnw\
+$(CIFIN): $(OBJS_DIR) cifin.c cifin-cmos26b.gen cifin.gen cifin.nw cifin.oldnw\
 	  cifin.others cifin.pw cifin-ami16.gen
 	rm -f $(CIFIN)
 	$(SC_CPP) -DSTANDARD cifin.c | ${SED_CMD} > $(CIFIN)
 
-$(CIFOUT): cifout.c cifout-cmos26b.gen cifout.gen cifout.nw\
+$(CIFOUT): $(OBJS_DIR) cifout.c cifout-cmos26b.gen cifout.gen cifout.nw\
 	   cifout.others cifout.pw cifout-ami16.gen
 	rm -f $(CIFOUT)
 	$(SC_CPP) -DSTANDARD cifout.c | ${SED_CMD} > $(CIFOUT)
 
-$(IBMCIFIN): cifin.c cifin-ibm.gen
+$(IBMCIFIN): $(OBJS_DIR) cifin.c cifin-ibm.gen
 	rm -f $(IBMCIFIN)
 	$(SC_CPP) -DIBM cifin.c | ${SED_CMD} > $(IBMCIFIN)
 
-$(IBMCIFOUT): cifout.c cifout-ibm.gen
+$(IBMCIFOUT): $(OBJS_DIR) cifout.c cifout-ibm.gen
 	rm -f $(IBMCIFOUT)
 	$(SC_CPP) -DIBM cifout.c | ${SED_CMD} > $(IBMCIFOUT)
 
-$(TMCIFIN): cifin.c cifin-cmos26b.gen
+$(TMCIFIN): $(OBJS_DIR) cifin.c cifin-cmos26b.gen
 	rm -f $(TMCIFIN)
 	$(SC_CPP) -DTIGHTMETAL cifin.c | ${SED_CMD} > $(TMCIFIN)
 
-$(TMCIFOUT): cifout.c cifout-cmos26b.gen
+$(TMCIFOUT): $(OBJS_DIR) cifout.c cifout-cmos26b.gen
 	rm -f $(TMCIFOUT)
 	$(SC_CPP) -DTIGHTMETAL cifout.c | ${SED_CMD} > $(TMCIFOUT)
 
-$(SUBCIFIN): cifin.c cifin-cmos26g.gen cifin-cmos14b.gen
+$(SUBCIFIN): $(OBJS_DIR) cifin.c cifin-cmos26g.gen cifin-cmos14b.gen
 	rm -f $(SUBCIFIN)
 	$(SC_CPP) -DSUBMICRON cifin.c | ${SED_CMD} > $(SUBCIFIN)
 
-$(SUBCIFOUT): cifout.c cifout-cmos26g.gen cifout-cmos14b.gen
+$(SUBCIFOUT): $(OBJS_DIR) cifout.c cifout-cmos26g.gen cifout-cmos14b.gen
 	rm -f $(SUBCIFOUT)
 	$(SC_CPP) -DSUBMICRON cifout.c | ${SED_CMD} > $(SUBCIFOUT)


dlmiles:
I do have a rework on this already at:

https://github.com/dlmiles/magic/tree/master-upstream-20241023-makefilev2  (76 commits for Makefile/build overhaul)

The specific commit that maybe most relevant might be https://github.com/RTimothyEdwards/magic/commit/39bb976736e1c1d34593355a156855def1c850b7

tskaar:
Yeah, seems like you're on the right track at least, from the GNU Make manual §5.7.3 Communicating Options to a Sub-make:

> The ‘-j’ option is a special case (see [Parallel Execution](https://www.gnu.org/software/make/manual/html_node/Parallel.html)). If you set it to some numeric value ‘N’ and your operating system supports it (most any UNIX system will; others typically won’t), the parent make and all the sub-makes will communicate to ensure that there are only ‘N’ jobs running at the same time between them all. Note that any job that is marked recursive (see [Instead of Executing Recipes](https://www.gnu.org/software/make/manual/html_node/Instead-of-Execution.html)) doesn’t count against the total jobs (otherwise we could get ‘N’ sub-makes running and have no slots left over for any real work!)
If your operating system doesn’t support the above communication, then no ‘-j’ is added to MAKEFLAGS, so that sub-makes run in non-parallel mode. If the ‘-j’ option were passed down to sub-makes you would get many more jobs running in parallel than you asked for. If you give ‘-j’ with no numeric argument, meaning to run as many jobs as possible in parallel, this is passed down, since multiple infinities are no more than one.

Doing something like:
`env -U MAKEFLAGS make -j$(nproc)`

Should be the _correct_ way, and it compiles/copies correctly as far as I can see

dlmiles:
It would be interesting to know, which operating system support this Makefile intercommunication, via what mechanism and what is a simple test to confirm it is working on a given platform/environment/setup.  Is the mechanism a inherited pipes/socket and custom environment variable ?

I am not sure I have ever seen it working on any platform in the way the documentation snippet indicates it should work, this is in effect sharing/borrowing worker process count, from a single global pool managed by the top-level $MAKE process.  Can't say I have seen that working on any platform, sounds really good if it does.


At the moment the design is on the basis:
* The top level superior $MAKE process gets to see the -j be it via command line or via $MAKEFLAGS.
* The subordinate $MAKE process do not get a -j because the option is forceabily removed by the empty string assignment to MAKEFLAGS.  I don't think there is a difference between setting to blank string (as now, thus negating any env inherited value) and unexport.  I would expect the observable outcome to be that the current method is safer (ensuring subdir Makefile are in effect -j1 and serialized).  It is possible if MAKEFLAGS exists, it is not overridden, but when MAKEFLAGS does not exist then $MAKE will construct one for the subordinate process allowing all inheritable options to be passed down by default.  Which is why the current setup is safer, as it appears to enforce serialization -j1 within all subordinate $MAKE processes which is what the design expects.
* The current project structure is not that bad (for total time to build metric) even in the situation when only the superior $MAKE process does parallel work, the sub-module nature helps farm out work and there is a high enough sub-module count that you get to use your cores.

---

To be clear on the status of this PR and this commit.

I have no objection to the commit in this PR, it does fix example snippet I show above https://github.com/RTimothyEdwards/magic/pull/424#issuecomment-3113922089  (when you have enough CPUs / speed), which is then confirmed by the build log snippet (if we assume $MAKEFLAGS is set giving a globally inherited -j to every $MAKE process, which is the equivalent to -j100 in my command snippet)

I am not against the idea of  improving the use case for infinite build processes, just pointing out there may still be other Makefile rework needed for that, because the project history is from a time of single CPU systems, so the Makefile's are constructed around that paradigm.

tskaar:
## For completeness

Using `MAKEFLAGS = ` in the top-level and with `MAKEFLAGS="-j16"` as environment gives the environment in `scmos/cif_templates/Makefile`:

`MAKEFLAGS=' -j16 --jobserver-auth=fifo:/tmp/GMfifo691404'`

While if I use `unexport MAKEFLAGS` in the top-level Makefile, I get:

`MAKEFLAGS=''`

The test case here was just adding a simple target to the Makefile that prints the MAKEFLAGS environment variable, i.e.:

```
all: ... print-env
...
print-env:
    @echo "MAKEFLAGS='$$MAKEFLAGS'"
```
So the latter case works correctly in my case, where the MAKEFLAGS environment flag is set, while in the former it has the previously mentioned racy beheaviour due to the sub-make processes inheriting the -j16 flag from the environment (as far as I understood).

---

> I am not against the idea of improving the use case for infinite build processes, just pointing out there may still be other Makefile rework needed for that, because the project history is from a time of single CPU systems, so the Makefile's are constructed around that paradigm.

Yeah, I agree, the PR forces the compilation to be OK, however it seems there is work to be done if one wants to use Make "correctly" in a highly parallel scenario.

dlmiles:
Thanks for the response, really useful feedback.

Please try `MAKEFLAGS += -j1` in the top level Makefile, see if you still get the socket arg but it doesn't edit the `-j1` setting.  An option like `-j16 --jobserver-auth=fifo:/tmp/path -j1` should work as expected, with `-j` taking priority.

The main thing is to serialize the (thread) demand from the subordinate $MAKE to asking for 1 thread only.
But the other features from 4.4.1 using the socket/fifo can stay and other inherited options can be present.

FWIW I am using 4.3 which was released 19-Jan-202.
The 4.4.1 released 26-Feb-2023 with the socket feature.

tskaar:
Different scenarios
---

Using `MAKEFLAGS += -j1` gives all _sub_-make processes use "-j1" as per make logs:

> make[1]: warning: -j1 forced in makefile: resetting jobserver mode.

While if I ran the top-level with "-j16", the top-level make still got "-j16" in the MAKEFLAGS, which sounds like what we want?

I also added to my MAKEFLAGS environment "-k" to see whether the sub-make processes kept that, and it did, where the scmos Makefile echo'd:

> MAKEFLAGS=k -j1

I.e., it got "k" from the environment and "-j1" from top-level Makefile.

Lastly, setting my MAKEFLAGS environment to "-j16" gave me at top-level Makefile:

> MAKEFLAGS=-j1

i.e. it overwrote what I had in my environment.

---

Summary:

1. `MAKEFLAGS += -j1` in top-level Makefile
  - Top-level make: Honors "-j" flag, but overrides MAKEFLAGS from environment
  - Sub-level make: Gets "-j1"
2. `unexport MAKEFLAGS` in top-level Makefile
  - Top-level make: Makeflags is not set anymore, not from flags nor environment
  - Sub-level make: Makeflags is not set anymore
3. Current scenario: `MAKEFLAGS = `
  - Top-level make: inherits from environment
  - Sub-level make: inherits from environment

dlmiles:
Thanks the the results.

Yes at this stage we're after -j16 to work toplevel only.  -j1 to be in effect for sublevels.

Maybe set the top level `MAKEFLAGS =` (as before) and add `MAKEFLAGS_SUBMAKE = -j1` then replace (at top level only) `${MAKE}` with `${MAKE} ${MAKEFLAGS_SUBMAKE}` (effectively adding the argument manually only to submakes) and this should be both compatible with 4.3 and 4.4+ and provide the expected parallelism with the expected build reliability.

With getting better parallel numbers, maybe if you try the tree https://github.com/dlmiles/magic/tree/master-upstream-20241023-makefilev2 (it should be possible to observe improvements with more CPUs available, between -j1 and -j16 when it is working as expected)

Maybe as each subdir is cleared to support Make 4.4 better, remove you'd remove the `${MAKEFLAGS_SUBMAKE}` option from that target in the toplevel Makefile, which will enabled it for 4.4+.

If you really need faster, take a look at `ccache` this may help out.  I believe this can be made available without modifying the project, via modification to envvars and/or toolchain resolution for 'gcc' etc...

tskaar:
Using your branch for reference:

`make -j1`:

```
________________________________________________________
Executed in   34.06 secs    fish           external
   usr time   21.40 secs  129.00 micros   21.40 secs
   sys time   12.57 secs  663.00 micros   12.57 secs

```

`make -j16`:

```
________________________________________________________
Executed in   15.98 secs    fish           external
   usr time   33.25 secs    0.00 micros   33.25 secs
   sys time   17.89 secs  757.00 micros   17.89 secs
```

So almost a 2x improvement there!

---

But yeah, I'd say it's fine with either honoring -j16 flag at top-level and then forcing -j1 in all sub-make processes at the moment.
I'd rather have it compile "correctly" rather than slightly faster :)

---

Just for reference compiling/packaging with the current PR, and just letting Make go ham on threads (with MAKEFLAGS="-j16" at environment level) I end up with:

```
________________________________________________________
Executed in   21.63 secs    fish           external
   usr time   31.10 secs    0.00 micros   31.10 secs
   sys time   15.05 secs  927.00 micros   15.05 secs
   
```
So roughly the same numbers as with the suggested `MAKEFLAGS += j1`  at top-level MAKE

RTimothyEdwards:
I assume that independently of the long conversation about parallelized compiles, I should be merging this pull request, because it sets the correct dependencies for the generation of the cifinput sections in the scmos tech file?

RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged to the opencircuitdesign.com repository.  Will be mirrored to github overnight.

PR #430, opened 2025-07-29
appimage v4

dlmiles:
Every version of recent appimage changes (v1,v2,v3) has worked in my GitHub repo (both when run on branch and on tag) but it keeps erroring with production releases (when refs/tags/8.3.xyz is pushed to GitHub) without providing any error message as to why.

So this will provide some additional log output to better see what various outputs are and better indicate progress between command around the problematic area. 

This will also ignore errors from a couple of commands and keep going.


RTimothyEdwards (approved):
Looks good to me.  I have noticed all of the github action failures;  they keep coming back to me as emails.

PR #433, opened 2025-07-30
appimage: allow git tags with annotation to match expected SHA for ChangeLog

dlmiles:
The situation is due to the difference for tags with message / annotation:

$ git tag -l -n1 8.3.534
8.3.534         "Tagging version 8.3.534"

$ git show-ref -s -d refs/tags/8.3.534
4426cc859e4f22ab96df941f123babd0d8076928
1ec8b6ee1a30c09debe5ad8a3208d810fdb1219f refs/tags/8.3.534^{}

$ git rev-list -n1 refs/tags/8.3.534
1ec8b6ee1a30c09debe5ad8a3208d810fdb1219f

Use of rev-lsit will always get the commit ID we expect to see for the workflow, so we allow either of the SHA above to match at the point of failure.


RTimothyEdwards:
Pulled and merged.

PR #436, opened 2025-08-05
8.3.536 paralleldevs

dlmiles:


dlmiles:
Just to add, removing the 'const' with a cast, is something that should probably have a `/* FIXME removes const */` comment.  Because it needs attention to come back later and evaluate.

The use of const in the `extflat/**` tree has shown up exactly which design issues need an alternative solution, the pinch points to having multi-threaded extraction.  If I ignore the tile change issues (TT_SIDE) and ignore the STDIO issues (this is more straightforward to fix).  Example of such are hidden in commits:

1dd0f32bcddf4991a0e059876f67fe91eb963779  REWORK cptr set to dterm->dterm_attrs but then may modify the data, under the newFmt feature from Feb2023  - need assistance to understand better

9d7545bd373301b87f94dc387e227ca20eb18496 REWORK EFVisit.c: efDevFixLW() dterm_attrs modified  - need assistance to understand better

b5b617126002f7046207a7cf8a28f064baef1f93 REWORK ext2spice: ext2spice.c: spcWriteParams() param_name modified l2/l3/etc..  - need assistance to understand better

f2c824c8ef52b11c324446e27e46016c26619de8 REWORK: EFHNConcatLook() modifies HierName in place, need thread-safe form  - have solution

706a2eea8debc4b131894e9a0cbde767b4937aaa REWORK: efBuildUse() modifies string in place, needs strndup() - have solution



At some point a CodeQL or similar AST based source analyser can be used to find locations where (removal of constness via a cast) occurs to cleanup, when such casts are necessary it should at least have a comment explaining why.

RTimothyEdwards (approved):
Got it, looks good, thanks.

RTimothyEdwards:
Pulled and merged.

PR #444, opened 2025-08-31
Fixed sed for bsd (macos) compatibility

### LINKED ISSUE #443: sed edit in place (-i) requires argument in bsd (macos)
The July 29, 2025 commit `bd541d3e6f1c1de2a984d750a3136e707a83cef5` prevents magic from compiling on macos.

The file [rules.mak](https://github.com/RTimothyEdwards/magic/blob/master/rules.mak) contains
```
${DEPEND_FILE}: ${DEPSRCS}
	${CC} ${CFLAGS} ${CPPFLAGS} ${DFLAGS} ${DEPEND_FLAG} ${DEPSRCS} > ${DEPEND_FILE}$$PPID.tmp
	${SED} -e "/#/D" -e "/ \//s/ \/.*\.h//" -e "/  \\\/D" -i ${DEPEND_FILE}$$PPID.tmp
	${MV} -f ${DEPEND_FILE}$$PPID.tmp ${DEPEND_FILE}
```

In BSD, the `-i` option requires an argument. Use '' for no backup.

d-m-bailey:
Fixes #443

### DIFF
--- a/rules.mak
+++ b/rules.mak
@@ -17,7 +17,7 @@ depend: ${DEPEND_FILE}
 # it indicates a missing dependency somewhere in a upstream/parent Makefile.
 ${DEPEND_FILE}: ${DEPSRCS}
 	${CC} ${CFLAGS} ${CPPFLAGS} ${DFLAGS} ${DEPEND_FLAG} ${DEPSRCS} > ${DEPEND_FILE}$$PPID.tmp
-	${SED} -e "/#/D" -e "/ \//s/ \/.*\.h//" -e "/  \\\/D" -i '' -- ${DEPEND_FILE}$$PPID.tmp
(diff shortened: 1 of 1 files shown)


dlmiles:
@d-m-bailey have you seen the updated MacOS instructions to install from brew gnu-sed ?

The recent autoconf detection to find GNU sed ?  When multiple sed dialects are installed on the same system.

All platforms (including FreeBSD, OpenBSD, SunOS/Solaris, CYGWIN, MacOS, have access to GNU sed).

A number of things I previously tried (in the limited time on this one) did not seem to work across that entire suite of dialects and sed argument syntax.


Maybe you can better describe the target OS, the version of sed, and the error seen.  Just to understand if this is something new or longstanding issue resolved with GNU sed and autoconf.

d-m-bailey:
Works if you follow the new instructions.

PR #473, opened 2025-11-23
defRead: Accept usename longer than 511 char

smunaut:
The previous code always read in a defined size buffer and any name longer gets cut which causes issue during extraction because any collision on the truncated name causes components to get "lost".

So here instead we keep a local stack buffer of 512 byte and use it if possible, but for longer string we allocate some space on the heap temporarily.

Some later error processing had to be refactored a bit to make sure we can always clean-up after ourselves once we're done with the buffer.

### DIFF
--- a/lef/defRead.c
+++ b/lef/defRead.c
@@ -2176,7 +2176,8 @@ DefReadComponents(
     Transform t;
     const char *token;
     char *dptr;
-    char usename[512];
+    char *usename, usename_buf[512];
+    int usename_len;
     int keyword, subkey;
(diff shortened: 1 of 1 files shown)


d-m-bailey:
@smunaut that was quick.

RTimothyEdwards (approved):
Looks good to me, thanks!

RTimothyEdwards:
Pulled and merged in the opencircuitdesign.com repository.  Magic version is now 8.3.578.  The github mirror will update overnight.

PR #435, opened 2025-07-31
Refactor: use functionality based checking instead of OS-level checking

tskaar:
What?
---

This refactors some of the code to use functionality based checking rather than OS-level checking for header includes

Motivation
---
For bleeding edge systems that are using GLIBC v2.42, `termio.h` has been removed, and `termios.h` is the replacement.
Due to recent refactors + removal of these headers one is not able to compile this when using GLIBC v2.24.

Note
---
Only tested the compilation on my machine (Arch Linux x86_64),
hopefully this doesn't break for other systems (BSD etc.)!

Closes: #434

PR #482, opened 2026-01-14
Add backward compatiblity code for <termio.h>.

jpc-lip6:
The backward compatibility code for <termio.h> has been removed starting from GLIBC 2.42. So it starts to fail on some bleeding edge distributions like openSUSE Tumbleweed. We took back the code removed from the GLIBC and put it back inside txInput.c.

### DIFF
--- a/textio/txInput.c
+++ b/textio/txInput.c
@@ -30,6 +30,29 @@ static char rcsid[] __attribute__ ((unused)) = "$Header: /usr/cvsroot/magic-8.0/
 #include <unistd.h>
 #include <ctype.h>
 #include <dirent.h>
+#ifdef __GLIBC__
+#  if (__GLIBC__ >= 2) && (__GLIBC_MINOR__ >= 42)
+
+/* Code borrowed from the GLIBC 2.41 */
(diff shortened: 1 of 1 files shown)


RTimothyEdwards:
Thanks for the input!

@dlmiles :  You have a good understanding of OS compatibility issues.  Any comment on the GLIBC changes?

jpc-lip6:
I confirm that your commit #435 works with the GLIBC 2.42.

When encountering those kind of deprecation problems, we have two ways of solving it:

* The quick way, when feasible and when you are not too familiar with the code you are patching, is to embed the compatibility code (as long as it stands).
* The *best* way, is to upgrade the code to newer standard. Which *may* incur a lot of work.

The question here may be why it was not merged into trunk ?

Yes, next time I will provides the build log. 

My builds, and their logs can be found here : https://build.opensuse.org/package/show/home:jpc-lip6/magic

(the faulty one are no longer there after I made the correction).

RTimothyEdwards:
This issue having been resolved in version 8.3.629, I am now closing this issue.

PR #505, opened 2026-05-01
Implement existing cell check for layout generation

Ahmed-Nematallah:
The current netlist_to_layout implementation lacks an existence check for subcircuits. When importing a SPICE netlist into an existing hierarchy, Magic's default behavior causes two critical failures:

Layout Corruption: While "paint" remains in place, all subcell instances are snapped to a default coordinate. This destroys the placement of any previously laid-out cells and causes a loss of instance-specific property modifications (e.g., transistor guard ring parameters).

Redundant Disk Writes: Cells already present in the search path (e.g., library cells or external IP) are treated as new, resulting in redundant .mag files being written to the current working directory, breaking synchronization with the source library (for incremental design).

Solution:
1. This patch introduces a pre-import verification step using load -fail to probe the search path for existing cell definitions.
2. It creates a mapping of existing vs. non-existing cells.
3. It gates the magic::generate_layout_add procedure, ensuring that only new cells are "filled" with components.

It explicitly excludes the top-level cell from the "skip" logic to allow for top-level connectivity updates while preserving the sub-hierarchy.

NOTE:
I'm not sure if checking "[info exists existing_cells($subname)]" in line 579 is fully needed. Please remove if unneeded.

Testing:
Verified on Magic 8.3.638. Incremental imports now successfully preserve existing placement and do not create local copies of library components.

This addresses issue #506

### DIFF
--- a/tcltk/toolkit.tcl
+++ b/tcltk/toolkit.tcl
@@ -536,6 +536,12 @@ proc magic::netlist_to_layout {netfile library} {
         if {[catch {load $subckt -fail -silent}] == 0} {
             puts stdout "Subcircuit $subckt successfully loaded."
             set existing_cells($subckt) "true"
+
+			# Make sure to load all cells into memory to avoid 
+			# corruption when saving all files (files get loaded
+			# without taking the scale into account)
+			select top cell
+            expand
         } else {
             puts stdout "Subcircuit $subckt not found. Will generate."
             set existing_cells($subckt) "false"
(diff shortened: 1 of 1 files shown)


RTimothyEdwards (approved):
This is a known issue, and thank you for providing the patch.
Normally I would use a "dict" type instead of an array, but this is fine.
Also normally I would use "catch" instead of testing for "info exists", but that's also fine.
Generally speaking, it looks okay to me, although I will run some tests.

RTimothyEdwards:
Pulled and merged to opencircuitdesign.com.  The github mirror will update overnight.

Issue #507, opened 2026-05-02
Toolbar Icon Refresh Failure on Layer Visibility Toggle


Ahmed-Nematallah:
1. Description
When a user attempts to make a layer visible via the toolbar (Left-Click), the layer correctly appears in the layout window, but the corresponding toolbar icon remains in the "hidden" (blank) state. Conversely, hiding a layer (Right-Click) works as expected, updating both the layout and the icon.
This also happens when using the "see" command, when using "see no metal1", it works, but "see metal1" does not update the toolbar, "see yes metal1" works.

NOTE: This bug is intermittent, but for my configuration, it happens nearly all the time.

2. Root Cause Analysis
The issue stems from a communication mismatch between the Magic C-core and the magic::toolupdate Tcl procedure.
Magic uses a callback mechanism to sync the GUI. When a "show" command is issued, the core often passes an empty string ("") for the optional third argument of the toolupdate command.
Current Code Logic: The procedure defines the third argument (layerlist) with a default value of "none". It then checks if {$layerlist == "none"} to determine if it needs to reassign variables to handle a shorthand command.
The Failure: Because the C-engine explicitly passes "", Tcl's default "none" is overwritten. The if check fails, the reassignment of $yesno to "yes" never happens, and the icon update logic is bypassed.

3. Steps to Reproduce
Open Magic with a technology file that includes a toolbar (e.g., Sky130).
Identify a visible layer and right-click its icon to hide it. (Icon turns blank, layer disappears, this is expected).
Left-click the same blank icon to show the layer.
Observe that the layer appears in the layout, but the icon remains blank (it is supposed to show up).

4. Proposed Fix
Redefine the toolupdate signature and internal check to recognize an empty string as the default indicator for shorthand commands. This ensures that even when the C-core passes "", the procedure correctly identifies the intent as a visibility "show" command.
Modified Logic:

Tcl
proc magic::toolupdate {win {yesno "yes"} {layerlist ""}} {
    ...
    if {$layerlist == ""} {
        set layerlist $yesno
        set yesno "yes"
    }
    ...
}

Environment
MX linux, TCL 8.6.13, magic 8.3.638 (latest version) compiled from source

RTimothyEdwards:
The PR has been merged and this issue should be able to be closed now.

PR #508, opened 2026-05-02
tcl/tk: Fix layer toolbar icon synchronization on left-click

Ahmed-Nematallah:
There is a bug where left-clicking a hidden layer in the toolbar successfully makes the layer visible in the layout but fails to update the toolbar icon to the "active" state.

The root cause is a positional argument mismatch in magic::toolupdate. When the Magic C-core issues a callback for a "see" command with two arguments, the third argument (layerlist) is passed as an empty string ("") rather than being omitted. This prevents the Tcl procedure from falling back to its hardcoded default of "none", causing the script to skip the logic that reassigns $yesno to "yes".

This patch changes the default value of $layerlist to "" and updates the conditional check to ensure the state and layer name are correctly reassigned regardless of how the C-core signals the update.

This pull request fixes Issue #507

### DIFF
--- a/tcltk/wrapper.tcl
+++ b/tcltk/wrapper.tcl
@@ -706,7 +706,7 @@ proc magic::cursorview {win} {
    *bypass logcommands resume
 }
 
-proc magic::toolupdate {win {yesno "yes"} {layerlist "none"}} {
+proc magic::toolupdate {win {yesno "yes"} {layerlist ""}} {
    global Winopts
 
    if {[magic::display] == "NULL"} {return}
@@ -728,7 +728,7 @@ proc magic::toolupdate {win {yesno "yes"} {layerlist "none"}} {
    # Don't do anything if toolbar is not present
    if { $Winopts(${topname},toolbar) == 0 } { return }
 
-   if {$layerlist == "none"} {
+   if {$layerlist == ""} {
 	set layerlist $yesno
 	set yesno "yes"
    }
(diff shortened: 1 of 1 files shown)


tcltk/wrapper.tcl
```diff
@@ -728,7 +728,7 @@ proc magic::toolupdate {win {yesno "yes"} {layerlist "none"}} {
    # Don't do anything if toolbar is not present
    if { $Winopts(${topname},toolbar) == 0 } { return }
 
-   if {$layerlist == "none"} {
+   if {$layerlist == ""} {
```
dlmiles: consider `if {$layerlist == "" || $layerlist == "none"}  {`

This seems more defensive programming, backwards compatible and manages both values in the same way.  Unless a blank string "" has a specific and independent meaning than "none" ?

dlmiles:
Feedback comments inline review.

tcltk/wrapper.tcl
Ahmed-Nematallah: That makes sense. In my testing with "see metal1", I observed an empty string being passed, which matches the default operands in most other Magic procedures. But it is understandable to keep backward compatibility.

(I did briefly wonder if someone might name a layer 'none', but that’s highly unlikely! 😬)

I guess the best approach is keeping the default argument as an empty string, checking for both, and adding a comment indicating you should not call a layer "none"

I'll update the PR.

Ahmed-Nematallah: Hi @dlmiles, I tested the new version and it works.

RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged to the opencircuitdesign.com database.  The github version will update overnight.

Issue #500, opened 2026-04-18
WASM Support --> NPM Package


Intubun:
Hi,
Ive recently been wondering wheter it would be possible to integrate magic into a wasm module, that could be hosted in an npm package. Im creating this issue just to have it listed here and that its being worked on. Ive created a fork of it, where gitub actions are already working and units tests (not yet in github actions as of now) are already passing. (Testing DRC, Extraction with parasitics, GDSII Export and CIF export). I would like to have it in the future as an npm package, so it could be integrated into various kinds of projects more easily.  I wanted to get guidance on how to best integrate it, so that it integrates nicely into the existing code base.
Thanks in advance for any feedback

RTimothyEdwards:
@dlmiles :  Any comment on this?  I know there is some kind of wasm compile in the github actions, and I also know that it has been failing on every single push I do.

dlmiles:
My main goal has been in maintaining the existing "WASM target still compiles" status as best as possible.  This does not necessarily mean it has been tested or used, but at least most of the battle is won in allowing consumers to feedback their experiences because they can get the project to build in the first place.

So would welcome testing framework to help feature coverage, even if the WASM target might not be considered a primary supported system (at this time), with Linux LTS desktop and MacOS being the primary supported system, but part of that problem is due to limited resources to support everything all at once so additional assistance is welcomed.

Need to see the commits first probably to provide better more specific feedback of what the original comment in this issue relates to.

The CI goals of keeping everything green and having some diversity (in build options) helps see around corners (or use cases, build environments) to provide a health indication as to how well the project compiles across different platforms.

---

On the specific NPM issue, a CI target can be created to build and publish to GitHub repository (which is NPM compatible), that might be the first step, as I believe it can be imported by NPM using a URL direct from GitHub (even if this is not at the official NPM packages repo).

I have a slight pet peeve with the current state of "package" systems, IMHO open source projects must:
* Allow forking and publishing of the artifacts into the forkers namespace (GitHub can do this via CI)
* This means the public CI and source code is auditable due to being 100% visible (the inputs, the method, the result)
* Artifacts published to 3rd party repos (such as NPM npmjs.com) should be digitally signed from the verbatim public CI output
* Details and verification mechanisms (for the digital signing) should is easily found to validate the authenticity of any published binary blob.  But more importantly the audit process should be as simple a forking the project and running the CI task to recreate the output.

This is why starting with the GitHub repository seems the place to start and the initial distribution point.

[...]

PR #509, opened 2026-05-05
WASM Support and NPM Package

Intubun:
Hi! This PR ports Magic to a headless WebAssembly target (as in issue #500 ) so it can run in Node.js, in the browser, or in a Web Worker — no X server, no Tk, not even a terminal needed. The artifact is published as the magic-vlsi-wasm package on npm and exposes a small JS API for running Magic commands and moving .mag, .gds, .cif, .ext, and SPICE files in and out of an in-memory filesystem.

Native builds are untouched on purpose. Every WASM-specific block is gated by EMSCRIPTEN or a configure flag, so a Tcl/Tk/X11 build comes out byte-identical to before. The simple_build_linux CI job is unchanged.

I've split the diff into seven topic commits so it's easy to review one piece at a time. The first three are useful on their own even if you don't care about WASM — they fix a handful of latent bugs and clean up some callback signatures that K&R-loose builds were quietly tolerating. The remaining four bring up the WASM target itself: the null graphics driver, the headless wiring, the Emscripten entry point and build, and finally the npm package with its CI workflows.
 @RTimothyEdwards @dlmiles

### DIFF
--- a/.github/workflows/main-aarch64.yml
+++ b/.github/workflows/main-aarch64.yml
@@ -1,14 +1,15 @@
-# This is a basic workflow to help you get started with Actions
+# CI for native ARM64 Linux build.
+# WASM is not built here — WebAssembly bytecode is architecture-independent,
+# so the single x86-64 WASM artifact produced by the CI workflow is identical
+# to what an ARM build would produce. See .github/workflows/main.yml.
 
 name: CI-aarch64
 
-# Controls when the workflow will run
 on:
   push:
   pull_request:
   workflow_dispatch:
 
-# A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   simple_build_linux_arm:
     runs-on: ubuntu-24.04-arm
@@ -23,39 +24,3 @@ jobs:
           ./configure
           make database/database.h
           make -j$(nproc)
-  simple_build_wasm_arm:
-    runs-on: ubuntu-24.04-arm
-    steps:
-      - uses: actions/checkout@v4
-      - name: Get Dependencies
-        run: |
-          git clone https://github.com/emscripten-core/emsdk.git
-          cd emsdk
-          ./emsdk install latest
-          ./emsdk activate latest
-      - name: Emscripten Diagnostic
-        run: |
-          source ./emsdk/emsdk_env.sh
-          echo "===== gcc -dM -E - ====="
-          echo | gcc -dM -E - | sort
-          echo "===== g++ -dM -E - ====="
-          echo | g++ -dM -E - | sort
-          echo "===== emcc -dM -E - ====="
-          echo | emcc -dM -E - | sort
-          echo "===== em++ -dM -E - ====="
-          echo | em++ -dM -E - | sort
-      - name: Build
-        run: |
-          source ./emsdk/emsdk_env.sh
-          # The --without and --disable in these build options is due to no WASM library being available for that feature
-          CFLAGS="--std=c17 -D_DEFAULT_SOURCE=1 -DEMSCRIPTEN=1 -g" emconfigure ./configure --without-cairo --without-opengl --without-x --without-tk --without-tcl --disable-readline --disable-compression --target=asmjs-unknown-emscripten
-          echo "===== defs.mak ====="
-          cat defs.mak
-          echo "===== defs.mak ====="
-          emmake make
-      - name: archive wasm bundle
-        uses: actions/upload-artifact@v4
-        with:
-          name: magic-wasm-bundle-arm
-          path: |
-            ${{ github.workspace }}/magic/magic.wasm
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,95 @@
+name: Publish npm package
+
+# Publishes magic-vlsi-wasm to npm.
+# Triggered automatically on version tags (v*); can also be run manually.
+# Requires an NPM_TOKEN repository secret for publishing.
+#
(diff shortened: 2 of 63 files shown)


Intubun:
I should add that the high number of lines only comes from a .tech file Im including as an example for having the Github CI Workflow running with test coverage, to detect issues with the build

dlmiles (commented):
Just some feedback here.

Please augment the ClientData insertion/change with comments indicating the data type of UNUSED status of the ClientData.

Please add the compiler/analyzer `/*ARGSUSED*/` before the function, this mitigates compilers/analyzer warning about unused arguments then the function prototype exists due to callback usage.

Here are some example snippets demonstrating such comments:

https://github.com/RTimothyEdwards/magic/blob/d0e38df6bf11fcc424a7352006152fca02cb9b2a/extcheck/extcheck.c#L166-L173

Forward declaration example (showing both a function where it is UNUSED and one where is has a used type):

https://github.com/RTimothyEdwards/magic/blob/d0e38df6bf11fcc424a7352006152fca02cb9b2a/ext2spice/ext2hier.c#L75-L76

The purpose of these is to help with static code analysis but also the casual observer reading can clearly see it is UNUSED and that intention has been checked by a human.

Don't worry about the @typedef custom attribute use I shall put these in later, this is to allow static analysis to validate the intended prototype.  Also to prevent the same function being used / claimed by two different prototypes.  Also to allow grep to find all callbacks of a particular type.

All these things also allow use of grep on the codebase to more easily find/audit things later.

This info mainly affected the commit https://github.com/RTimothyEdwards/magic/pull/509/commits/58174e2a97ea2d869b4e82c57ba6add85692ff02


Thanks

npm/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "magic-vlsi-wasm",
+  "version": "8.3.637",
```
dlmiles, line 3: This needs to be modified via automation.  Maybe it should follow the AppImage form to include the date of the commit-id, the commit id, like `8.3.637~20260414~d157eea` this conveys more info to the user, confirm the source origin, but also allows an updated release to be made against 8.3.637 independent of later versions like 8.3.638.  For example as a security update for a long-term-support.

.github/workflows/main-aarch64.yml
```diff
@@ -23,39 +24,3 @@ jobs:
...
-    runs-on: ubuntu-24.04-arm
-    steps:
-      - uses: actions/checkout@v4
-      - name: Get Dependencies
-        run: |
-          git clone https://github.com/emscripten-core/emsdk.git
-          cd emsdk
-          ./emsdk install latest
-          ./emsdk activate latest
-      - name: Emscripten Diagnostic
```
dlmiles, line 36: Please keep this diagnostic section in the (new/moved) GitHub CI  they provide diagnostic to fix the CI as all the info is in the logs to see, you don't have to spend more hours working out the problem.  Things (emsdk) change over time.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,95 @@
...
+          node-version: '22'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Install emsdk
+        env:
+          # --- emsdk version switch -----------------------------------------
+          # Pinned: must match .github/workflows/main.yml. See post-build.sh.
+          # To test against the latest emsdk, comment the pinned line and
+          # uncomment the "latest" line below.
+          EMSDK_VERSION: '3.1.56'
```
dlmiles, line 45: Can this be a CI workflow argument, with the default value you need.  Please link issue or one-line explain the purpose for locking a specific version.  The answer to the "Why"
It is useful for the CI to also build against 'latest' even if that is not the one published, the purpose it to keep an eye on when it breaks.
Allowing manually running of GitHub CI (with `on workflow_dispatch:` would allow overriding the value and trying different versions and coming back later to see if it works.

https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,95 @@
...
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v5
+        with:
+          node-version: '22'
+          registry-url: 'https://registry.npmjs.org'
```
dlmiles, line 37: Please publish to GitHub first.  This would be: `registry-url: 'https://npm.pkg.github.com`

It does not make sense to publish to NPM registry until it has at least been checked over, and you can still download from the project package from GitHub using NPM in the short term.

This would allow ironing out issues, testing, evaluation of the package.  Until everyone is happy it is ready for a more public release.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,95 @@
...
+
+# actions/upload-artifact@v5 still runs on Node.js 20. Force Node 24 to
+# silence the deprecation warning until upload-artifact ships a Node-24
+# release. Drop this once upgraded.
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
+jobs:
+  build-and-publish:
+    runs-on: ubuntu-22.04
```
dlmiles, line 28: I think this was copied from the simple_linux_wasm but not from the arm64 that uses 24.04.

Can we update this (to `24.04` at least)  ?  any reason not to use `latest` ?

WASM is still somewhat bleeding edge and I would guess benefits most from using `latest` ?

dlmiles (commented):
Feedback on this commit.

PR #510, opened 2026-05-06
WASM Support, NPM package

Intubun:
Ive had to force push something into my repo, so I couldnt reopen that PR, so heres a new one.

Ive implemented what you asked for:
- Keep diagnostics section in main-aarch64.yml (Btw. why in the aarch64? WASM should be indipendent of that?)
- automatic versioning in package.json
- npm.pkg.github.com as preliminary publish desitnation
- ubuntu and emsdk version changed to latest and emsdk version selectable, if run manually
- Added the comments for static code analysis


Intubun:
@dlmiles

RTimothyEdwards (approved):
I have a few issues with the scripts used for the CI, but nothing major (a few methods are outdated with respect to recent code changes, and I feel like it would make more sense to use the existing scmos.tech technology file than to import the siliwiz.tech technology file).

This appears to compile the wasm version without Tcl support?  I'm concerned that that removes a lot of capability from the tool.  But if I've done my job maintaining backwards compatibility with the non-Tcl build, then it should work okay.

I will approve this but would like to hear from Darryl Miles.

dlmiles (commented):
I think my previous instruction was misunderstood slightly.

The main-aarach64.yml simple_build_wasm_arm can be removed, if there is a new replacement, that does the same and optionally publishes a package.   It is only the "Emscripten Diagnostic" step that needs to be copied into the new WASM CI build task.  Which was the matter in my previous instruction.

It is OK to remove the simple_build_wasm_arm step, as shown in the previous PR/change-set iteration.

I assume it should not matter if Intel or Arm was used to build WASM the output binary blob should the same ?

The same is true of removing the simple_build_wasm from main.yml (so the WASM build is in a new workflow)

Intubun:
Ive changed the example to a scmos example. Also Ive updated the readme in /npm, since it was outdated.

Yes, it doenst use tcl at the moment, but there is a parser that can take a script and run it. No complex logic as of yet.

@dlmiles I was confused, since the comment was in the aarch file. Yeah WASM / Emcripten does not care if it is run and/or compiled on different ISAs. (Could even run on a more modern phone probably). Give me a minute and I make a change.

dlmiles:
Maybe the `npm-publish.yml` is better called `main-wasm.yml` as the main goal is to always build a WASM version.  It can then optionally also publish to a package repo.

Intubun:
@dlmiles Ive updated as u requested. I hope its satifiying. tcl would be the next todo item for me, but what im working on can now continue forwards. Btw. absolute thanks to all of the work that has been done by both of you (and all of the other maintainers)

dlmiles:
> I have a few issues with the scripts used for the CI, but nothing major (a few methods are outdated with respect to recent code changes, and I feel like it would make more sense to use the existing scmos.tech technology file than to 
import the siliwiz.tech technology file).

Please cite the methods of concern, otherwise we're guessing, so at least people can take a look.


> This appears to compile the wasm version without Tcl support? I'm concerned that that removes a lot of capability from the tool. But if I've done my job maintaining backwards compatibility with the non-Tcl build, then it should work okay.

There has never been Tcl support in the WASM build.   This requires building Tcl for WASM object target, then using the output object to link magic against.  The problem here is the Tcl engine assumes a Unix process environment at runtime and is based on 1980 programming methods and conventions, threading an after thought, large scale IO after thought (solved ~15y later).  I believe it is possible to have a special cut-down Tcl core that might help towards this goal.

The current WASM support presented in this change-set is no different to how it has always been.  So this change-set do not change the current situation and is not a backward step in this regard.

Now on this subject (of Tcl language support for byte-code runtimes).  I have as a separate and independent project a working Tcl engine that is specifically designed to run on byte-code runtimes.  It still has matters to improve that I am working on over time, over 1000 unit tests that run Tcl script snippets and compare directly against Tcl8.6 and Tcl9.0.  At this point it is just parser/ast-generation/interpreted-execution-engine/runtime-library, I am just starting to look at native Tcl bytecode support (yes Tcl has an internal bytecode).  Then from getting this to work I have a template for targeting other byte-codes directly such as WASM.  I expect many months before announcing something viable.

This approach is to bring Tcl up to par with other language support in the byte-code space (as it is somewhat forgotten/left-out by the wider community), but once up to par first-class Python support arrives for free (which is the language the EDA community wants).  So now there can be a REPL/shell-interface that understands both Python and Tcl at the same time.

My goal here is the best cross-language interop on unconstrained desktop environment, you could say targeting utility, productivity and performance.  WASM in the browser-constrained environment running headless magic is useful, but my interest is more in the interop interfaces that allow that to happen naturally as by-product of separation at the right internal boundaries.  Luckly most of those boundaries already exist just a few things need unpicking and putting in the correct place.
 
WASM at this time is still receiving major features every 3 months (as the details are worked out) with browser 3 months behind that, in some cases these features might be considered minimum requirements in 2026 but the details take a while to work out, so their process seems to be steady stable and well thought out.

> I will approve this but would like to hear from Darryl Miles.

Still looking over things, needs to clone / merge run GitHub CI and try somethings locally, so will take least this week and weekend.  So far have I have only commented in the things I can see from GH.

RTimothyEdwards:
@dlmiles :  Nice!  So, based on your comments, I think that I have no particular objections to this pull request once you've run your tests and approve it yourself.

As for the scripts, now that I've looked over them more carefully, three of them (DRC, GDS, and CIF) are trivially simple (three-line scripts).  These three scripts look fine.  In the "extract" script (extract.tcl):

line 20:  Should be "select top cell", which will always select the full layout.
line 21:  The preferred usage replaces "extresist all" here with "extract do resistance" before line 17.
line 25:  "ext2spice rthresh 0" does absolutely nothing in this context.

It's not clear to me what design is being passed to the extract script, but only a layout that is not hierarchical will produce a valid output.

Intubun:
under the /npm/examples is min.mag (a BJT Transistor in the scmos technology). Thats the basis for all the example scripts

RTimothyEdwards:
@dlmiles :  I have no additional comments here.  If you agree, I'll go ahead and merge this pull request.

dlmiles:
Confirm no further changes from another walk through.

Was hoping to have had time  to put up on github and test build releases and confirm release mechanism and how the final binary looks but fixups can be done after merge.

dlmiles (approved):

RTimothyEdwards:
Pulled and merged on opencircuitdesign.com.  The github mirror will update overnight.

PR #516, opened 2026-05-17
Increased maximum stipple count from 29 to 255 & remove duplicate file

Ahmed-Nematallah:
### The Problem
The PostScript plotter in Magic contains a bottleneck where the stipple capacity limit is hardcoded to 29. If a user attempts to use a PDK that exceeds this number, the PostScript engine crashes. This particularly a problem as mos.24bit.dstyle has 36 stipples out of the box.

Additionally, the stipple indexing utilized a base-30 encoding scheme which is unfamiliar to most people.

The Solution
This PR:
1) Modifies the postscript header to allow for up to 255 stipples (the upper bound of the font-encoding array approach).
2) Replaces the base-30 encoding with straightforward base-10 calculation.
3) Found and deleted a duplicate, unused copy of the PostScript template file in the tktcl directory.

Architectural Discussion / Proposal for Maintainers

- While reviewing the codebase to implement this fix, I noticed that the PostScript prolog execution environment is tightly coupled and "interlocked" across two separate locations:
- Some structural envelope closures (such as certain begin blocks and the trailing MAGICsave restore / %%EOF markers) are hardcoded directly into the compiled C binary via fprintf statements.
- The core dictionary structures and macros (like /dp) live in this external template file.
- Because these files share an interlocking dependency, a user cannot safely customize the external template file without risking a structure mismatch with the compiled C binary. Since changing one often forces a modification and recompilation of both anyway, would the maintainers be open to a follow-up PR that fully consolidates this external file directly into the C program as a static literal string?
- Consolidating it would clean up the repository architecture, eliminate an unnecessary runtime disk-read dependency, and keep the tightly coupled PostScript environment safely unified in a single source file.


RTimothyEdwards (approved):
Looks good to me.  I didn't even know there was a redundant (and slightly outdated) copy of magicps.pro in tcltk/.  I have no idea what it was doing there.

I assume I had some reason for the base-30 encoding, but I cannot remember what it is, and the base-10 encoding seems to make more sense.

RTimothyEdwards:
Pulled and merged to the repository on opencircuitdesign.com.  The github mirror will update overnight.

PR #519, opened 2026-05-22
Add TCL support for WASM builds

Intubun:
Hey everyone,
ive been working on adding tcl support to the magic-vlsi-wasm package. Ive added:

- A WASM build variant alongside existing non-TCL build
- tcl/tk gets compiled as static library and linked directly into magic.wasm
- new --with-tk=no configure path for TCL without Tk builds, since WASM only runs headless
- CI workflow for all these changes
- More memory by standard (64 MB for heap and 10 MB for Stack, can be adjusted at runtime)

Matrix and all CI already ran on my fork and everything is green.


npm/build.sh
```diff
@@ -76,53 +93,124 @@ sed_strip_cr() {
...
+  fi
+  rm -f defs.mak database/database.h
+
+  # Strip Windows CRLF line endings (no-op on Linux-native files).
+  sed_strip_cr configure
+  find scripts/ -type f -print0 | while IFS= read -r -d '' f; do sed_strip_cr "$f"; done
+
+  if [ "$variant" = "tcl" ]; then
+    ensure_tcl_built
+    CFLAGS="--std=c17 -D_DEFAULT_SOURCE=1 -DEMSCRIPTEN=1${EXTRA_CFLAGS}" \
```
dlmiles: -DEMSCRIPTEN=1${EXTRA_CFLAGS}

Is this correct without a space before the dollar ?

There is at least one more site below (please check all the files in npm/**).  This may appear to work when EXTRA=CFLAGS="" but if you configure it to a non-blank string I would expect things to error.

npm/README.md
```diff
@@ -122,14 +166,35 @@ If you want to rebuild the WASM module yourself, see
...
+### TCL variant: cloning the TCL source tree
+
+The TCL variant links against a static WASM build of
+[tcltk/tcl](https://github.com/tcltk/tcl), pinned to a specific commit in
+[`npm/tcl.ref`](tcl.ref). `build.sh` clones the source tree automatically
+into a `../tcl` sibling directory on the first run:
+
+```bash
+# Override the default clone location
+TCL_REPO=/path/to/existing/tcl bash npm/build.sh --variant=tcl
```
dlmiles: Not clear what the function of this part is.  If the WASM TCL is built, you would expect the C TCL to be installed as well.  So there is a $prefix/lib/tclConfig.sh which should contain all the information a dependant project needs to consume Tcl.  This config pre-dates pkg-config system that, but serves the same purpose.

If this part is for making the Tcl source available to build the WASM target.  This is probably best considered being done entirely inside Tcl.  For example the 'unix' subdir is for UNIX targets, the 'win' subdir is for Win32 targets, etc...  I'm sure Tcl upstream project would welcome having a contribution with a 'wasm' subdir that builts Tcl for a WASM target, possibly with access to WASI (which is the closest APIs suite that should cover core functionality).

When maybe a shim for WASM in the browser.  To manage WASI and non-WASI and this would be a more minor target adjustment.

This comment should not hold up the contribution here, just feedback that access to a working Tcl on WASM target might be something more Tcl users want and sits well with the construction of the project in a wasm/ subdir.

npm/tcl.ref
```diff
@@ -0,0 +1,12 @@
...
+# Format: shell-style "VAR=VALUE" lines (no spaces around =).
+# Lines starting with # or blank lines are ignored.
+#
+# To take a newer TCL release into magic-wasm:
+#   1. Update TCL_REF below to the desired commit SHA (or tag/branch).
+#   2. Bump magic/VERSION as usual.
+#   3. Commit + push. CI rebuilds and republishes.
+
+TCL_REPO_URL=https://github.com/tcltk/tcl.git
+TCL_REF=84b23291b0dd811d642abef4ec7a55473c3eccb3
```
dlmiles: TCL_REF=main

Is there a reason this is pinned by default ?   Ideally should this not take the latest tag of the stable branch.  So by default CI will also rebuild against newer TCL.

Can the build use the environment $TCL_REF in priority to the default here, "refs/tags/9.0.3"

Can the build use the environment $TCL_REPO_URL in priority to this default here.

Can release notes be constructed around the GitHub artifact creation, so at the bottom of the workflow task.  The goal is to at a glance see the versions of important things that were use by the CI.

This might be compiler, emscripten, tcl commit/tag/url.

Ideally the CI should be optimistically building for newer current stuff in low maintenance mode by default, so no one need to update the SHA here.

Maybe in this case when the GitHub workflow runs, one task is a form of information gathering, maybe to decide the TCL_REF value and report in summary the `git log -n1` style output.  It then sets $TCL_REF in the workflow, to this value here can be set to `TCL_REF=main`.  The information gathering is then used to render a $GITHUB_STEP_SUMMARY to display things with the *.wasm downloads.

If the CI requires pinning and heuristics over picking a working version to use, the pinning would be done in the workflow file.

This would then extend to using on:workflow_dispatch to specify a specify TCL_REF.

All this is to help in 6 months when it breaks to have as much information as possible available in CI to spend the least amount of time rebuilding magic using the previously working TCL_REF to confirm the problem is with the current state of the upstream project, not with magic's recent changes.

Part of the reason why the CI workflow default TCL_REF should be computed on the most recent stable branch `main` release tag, is only use Tcl versions that the Tcl project consider release ready.

If you are locking a version (because that is needed) around the SHA please show the `git log -n1` info nearby, to show the commit vital statistics (date/tag/author/comment).

dlmiles (commented):
A quick skim through for some initial feedback.

npm/build.sh
Intubun: I had the spaces at other places and therefore it compiled correctly, but yeah, Ill fix it.

dlmiles:
Can we resolve the Semantic Version scheme for NPM package versions as well ?

Comment added after close https://github.com/RTimothyEdwards/magic/pull/512#issuecomment-4448268370

Specifically what is bad is the use of a raw git hash after a decimal point such as example: `8.3.799-20260622.01234cdef`.  This as I understand it, is the scheme used in the first release.

It is better to be `8.3.799.20260622git0123cdef` where `git` seems to be the defacto standard in the NPM package world for this usage?  as it conveys what the sequence of the following characters means ?  if you recall the goal is to allow insertion of security updates (if the project wanted to) after the events in the future.  The goal would be some one specifying they want `magic==8.3.799` would get the latest version in the `8.3.799` series, thus picking up an inserted security update.  Each part after decimal point should follow an order (as per semantic versioning), but use of the githash as `.01234cdef` breaks this as the next hash might be `.0fedc321` but that now appears to be before.

Not thinking about this in advance, makes it impossible to fix in the future.
In practice the project may just request users to upgrade to the latest release, but having the option is always best planned for in advance.

I just have not done to research (myself) to see how other NPM projects manage including the git-hash with semantic versioning.  But I know that they do manage this within their visioning scheme.   But what I can see at this time is exactly what I didn't want.

Intubun:
There are two distinct version formats in this workflow worth separating:

Published releases are tag-driven: pushing git tag v8.3.799 causes the workflow to publish exactly 8.3.799 to GitHub Packages — clean three-component semver, no suffix. Security patches would be published as 8.3.800, 8.3.801 etc. Users who declare ~8.3.799 in their package.json receive those automatically on npm update. This is the standard npm mechanism for the use case described and is already in place.

CI snapshots (non-tag builds) are packed as artifacts for local inspection but are never published to GitHub Packages. The format has been updated from 8.3.799-20260622.01234cdef to 8.3.799+20260622.git01234cdef. The original - was the real problem: in semver it marks a pre-release, meaning the snapshot sorted before 8.3.799 rather than after. The + separator correctly marks it as build metadata ("this is 8.3.799 built from this commit"), and the git prefix makes the hash immediately recognisable. Since snapshots are never published, the hex ordering concern does not arise in practice.

Regarding 8.3.799.20260622git0123cdef — the intent is right but npm's node-semver strictly requires exactly three numeric components and rejects a fourth, so this format cannot be used regardless of convention.

npm/tcl.ref
Intubun: Ive deleted the tcl.ref and now the tcl verison is tagged to the latest stable release. You can override the verion with a manual workflow dispatch. Also a build summary was added.

dlmiles:
The upgrade to latest is always an option, and not the point I am drawing attention to (as it is an obvious matter).

The security update means keep the same 8.3.799 version but receive a security patch on top (the version needs to be inserted into the range).  For example there maybe users who lock a specific version of magic because it has been tested/checked and because they are building chips (with a 1 year lifecycle and expensive stability concerns) who want stability not new features, so they would use `magic==8.3.799` because it passed their internal processes at some point in the past.

The idea of a security (or major bugfix/correction) patch issue in this way is to insert it into the range 8.3.799 range after 8.3.800 was release (retrospectively).

Is it allowed in semantic versioning to have `8.3.799.1+20260622git0123cdef` and will that version always be considered a match for `magic==8.3.799` when the original release was `8.3.799+20260622git0123cdef`.  By using the `+` it seems like that manages the extra part, and by not using `.` we can still have a 4th part to the version to insert the security update ?

---

Ah then if it strictly only allows 3 parts to a version, I am saying the main release also need the `+20260622git0123cdef` part to allow the security/major-fix to be inserted.

Intubun:
npm does treat any three part sementic versioning the same, no matter the extension. npm, correct me if im wrong, does not have any way to retroactively insert security patches. The only path for patching is to increase the version number. The part after that is only for traceability. At least thats my knowledge on how the npm registry handles it.

dlmiles:
I shall research and test some things over the new few days.

Is there any reason why `${base}-${date}+git${hash}` is a bad idea ?  This would seem to address all constraints:
* NPM semver works, as 3 parts, a `+` introduces supplementary information.
* An update for the product version `8.3.799` can be inserted, as `799-20260522` is older than `799-20260801`
* The release build would use versioning in the form `8.3.799-20260522` due to this.
* Also `${base}~${date}+git${hash}` would be ok, if in NPM world use if tilde '~' is a defaco standard.

Unclear if `magic>=8.2.799` matches.

As I understand it from your comments: a `.` after a `+` has a different meaning (it is not part of semver matching), but in the above scheme is was removed anyway, side-stepping the concern, but in NPM and the casual observer (who does not know/understand semver rules).

I don't see having the date on display being a bad thing, a human reading can immediately make a judgement call from seeing a date on they should consider to update, based around all the information they know from their view of the world.

Intubun:
yeah semver matching is different depending on the symbol. ".", "-" and "+" have different meanings. "." is for sectioning the main version, "-" can be used to declare alpha / beta version. "1.1.1-alpha.1" for example. and "+" only adds additional information, for example build time information, etc. But I should also do more research on the topic

Intubun:
I did my research, the "-" has to be actively selected by the user, since its only intended for prerelease version. Its mostly intended to implement patches by bumping the last part of the version. for example: magic>=8.3.799 would not match with 8.3.799-20260526, since its considered a prerelease version of a package. So Users would not get the update, unless they actively pin that version with magic@8.3.799-20260526, but it wont update the package automatically with the next npm install / npm update. And since @RTimothyEdwards magic repo is mirrored by github every day at around 9am my time, there is a more or less daily release schedule and if we ship it just as 8.3.799, then all the users would get the newest version, if they run npm install or update.

dlmiles:
My research gets to here.

Using a version in the form: `8.3.799020261231+git0123cdef`

The inserted ZERO just makes the date easier to read.  Can also be used for .1 to .9 style 4th part, the point is the scheme leaves the option open to decide another day.  Which is what thinking about it now is all about (creating future options).

Version locking can be done in a number of ways, see below. `<8.3.800xxxxxxxxx`

```
VALID 8.3.799~20261231+git0123cdef -> false
VALID 8.3.799_20261231+git0123cdef -> false
VALID 8.3.799$20261231+git0123cdef -> false
VALID 8.3.799%20261231+git0123cdef -> false
VALID 8.3.799^20261231+git0123cdef -> false
VALID 8.3.799&20261231+git0123cdef -> false
VALID 8.3.799*20261231+git0123cdef -> false
VALID 8.3.799(20261231+git0123cdef -> false
VALID 8.3.799)20261231+git0123cdef -> false
VALID 8.3.799=20261231+git0123cdef -> false
VALID 8.3.799[20261231+git0123cdef -> false
VALID 8.3.799]20261231+git0123cdef -> false
VALID 8.3.799{20261231+git0123cdef -> false
VALID 8.3.799}20261231+git0123cdef -> false
VALID 8.3.799#20261231+git0123cdef -> false
VALID 8.3.799@20261231+git0123cdef -> false
VALID 8.3.799:20261231+git0123cdef -> false
VALID 8.3.799;20261231+git0123cdef -> false
VALID 8.3.799.20261231+git0123cdef -> false
VALID 8.3.799,20261231+git0123cdef -> false
VALID 8.3.799>20261231+git0123cdef -> false
VALID 8.3.799<20261231+git0123cdef -> false
VALID 8.3.799/20261231+git0123cdef -> false
VALID 8.3.799?20261231+git0123cdef -> false
VALID 8.3.799|20261231+git0123cdef -> false
VALID 8.3.799020261231+git0123cdef -> true
VALID 8.3.799020261231+git0123cdef -> true
VALID 8.3.999020261231+git0123cdef -> true
VALID 8.3.1000020261231+git0123cdef -> true
VALID 8.3.1001020261231+git0123cdef -> true
COMPARE 8.3.799 vs 8.3.799020261231+git0123cdef -> -1 BELOW
COMPARE 8.3.800 vs 8.3.799020261231+git0123cdef -> -1 BELOW
COMPARE 8.3.799020261231+git01230000 vs 8.3.799020261231+git0123cdef -> 0 EQUAL
COMPARE 8.3.799020261231+gitffff0000 vs 8.3.799020261231+git0123cdef -> 0 EQUAL
COMPARE 8.3.799020260831+git0123cdef vs 8.3.799020261231+git0123cdef -> -1 BELOW
COMPARE 8.3.799020270101+git0123cdef vs 8.3.799020261231+git0123cdef -> 1 ABOVE
COMPARE 8.3.999020270101+git0123cdef vs 8.3.799020261231+git0123cdef -> 1 ABOVE
COMPARE 8.3.1000020270101+git0123cdef vs 8.3.799020261231+git0123cdef -> 1 ABOVE
COMPARE 8.3.1001020270101+git0123cdef vs 8.3.799020261231+git0123cdef -> 1 ABOVE
MATCH version=8.3.799020261231+git0123cdef range=~8.3.799 -> true
MATCH version=8.3.800020261231+git0123cdef range=~8.3.799 -> true
MATCH version=8.3.1000020261231+git0123cdef range=~8.3.799 -> true
MATCH version=8.3.1001020261231+git0123cdef range=~8.3.799 -> true
MATCH version=8.3.799020261231+git0123cdef range=~8.3.7990 -> true
MATCH version=8.3.800020261231+git0123cdef range=~8.3.7990 -> true
MATCH version=8.3.1000020261231+git0123cdef range=~8.3.7990 -> true
MATCH version=8.3.1001020261231+git0123cdef range=~8.3.7990 -> true
MATCH version=8.3.799020271201+git0123cdef range=~8.3.799 -> true
MATCH version=8.3.799020270101+git0123cdef range=~8.3.799 -> true
MATCH version=8.3.799020270202+git0123cdef range=~8.3.799 -> true
MATCH version=8.3.799020270303+git0123cdef range=~8.3.799 -> true
MATCH version=8.3.800020270101+git0123cdef range=<=8.3.799900000000 -> false
MATCH version=8.3.801020270202+git0123cdef range=<=8.3.799900000000 -> false
MATCH version=8.3.801020270202+git0123cdef range=<=8.3.7999x0000000 -> false
MATCH version=8.3.801020270202+git0123cdef range=<=8.3.7999xxxxxxxx -> false
MATCH version=8.3.799020271201+git0123cdef range=~8.3.7990 -> true
MATCH version=8.3.799020270101+git0123cdef range=~8.3.7990 -> true
MATCH version=8.3.799020270202+git0123cdef range=~8.3.7990 -> true
MATCH version=8.3.799020270303+git0123cdef range=~8.3.7990 -> true
MATCH version=8.3.800020270101+git0123cdef range=<8.3.8000xxxxxxxx -> false
MATCH version=8.3.801020270202+git0123cdef range=<8.3.8000xxxxxxxx -> false
MATCH version=8.3.801020270202+git0123cdef range=<8.3.800xxxxxxxxx -> false
```

Intubun:
Agreed. Let me implement it. It maintains correct orderning, while allowing small patched to be implemented on a as needed basis. Havent thought about that.

npm/README.md
Intubun: it was a local override, in case there was a local version, for example for dev on the tcl branch. however i removed it after careful consideration in commit [16c0ab3](https://github.com/RTimothyEdwards/magic/pull/519/commits/16c0ab35b1fcf07092f48ae5decf0ec4f5bf33d6) So now only with override of the version and repo to be cloned.

Intubun:
@dlmiles Is there anything left to do from your point of view?

dlmiles:
I'll update tomorrow on this (not expected to find any concerns).  Have got it locally built to look at just not tried it out.

@RTimothyEdwards any feedback from your side ?

RTimothyEdwards:
@dlmiles :  No comments from me.  All the WASM stuff is entirely outside my area of expertise and I am relying on you to give it the sign-off.

npm/build.sh
```diff
@@ -76,53 +93,119 @@ sed_strip_cr() {
...
+  tcl|notcl) build_variant "$OPT_VARIANT" ;;
+  both)      build_variant notcl
+             build_variant tcl ;;
+esac
 
 # --- optional test -----------------------------------------------------------
-# Runs the same smoke test that CI runs (see .github/workflows/main.yml).
+# Runs the same smoke test that CI runs (see .github/workflows/main-wasm.yml).
 if [ $OPT_TEST -eq 1 ]; then
   cd "$SCRIPT_DIR"
```
dlmiles: This use of `cd` wants to be in a subshell `(cd dir; npm command)` so the `pack.sh` below line at 130 has a chance to run.

npm/examples/extract-tcl.js
```diff
@@ -0,0 +1,39 @@
...
+import { writeFileSync, mkdirSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { resolve } from 'node:path';
+
+export async function run({ magFile = DEFAULT_MAG, tech = DEFAULT_TECH, outputDir = DEFAULT_OUT } = {}) {
+  const { magic } = await createMagic();
+  const { FS } = magic;
+  const { tech: techName, cell } = loadCell(FS, tech, magFile);
+
+  magic.runScript(loadScript('extract-tcl.tcl', techName, cell));
```
dlmiles, line 13: These example scripts need to wrap and throw runtime exceptions so that there is a readable diagnostic output when they fail.  This does not seems to happen by default, not at least if you run as `node examples/extract.js`

They ideally need to run as many individually as possible, as in the first failure should not stop all other scripts from being processed, unless it is known to be dependant upon the output.  It is still useful to know when 5 of 7 tests were successful.

This make them more useful and less cryptic in their use as a CI/fitness testing function.

Example (all bet it not a good one):
```
try {
  // Thing that can fail (ideally this might be the whole function not just runScript, if used for CI tests)
  magic.runScript(....);
} catch(e) {
   console.error(e.toString()); // ensure CI outputs emits error details
   console.error(e.stack?.toString());
   throw e; // ensure caller sees failure
}
```

dlmiles (commented):
Just sending this now for feedback been in draft for over 24h.

Waiting on https://github.com/RTimothyEdwards/magic/issues/523 to continue testing/evaluation as it breaks WASM test at this time (see recent CI/Actions).

dlmiles:
I would expect the try/catch to show the nature of the failure in CI such as at https://github.com/RTimothyEdwards/magic/actions/runs/27055629781/job/79859362144#step:8:38  (at the moment it doesn't provide any feedback, which is really useful information).

Ideally the failure data (stdout/stderr) should be searched for something that looks like a stack trace, then it can augment, this is the standard looking output, note I don't see stdout/stdserr in the CI logs, which I think are useful to expose, ensure to build with -g and link with -gsource-map and copy `magic.wasm.map` alongside, I also executed with `--enable-source-maps` not entirely sure what is a good set of runtime options to use in CI to elicit maximum feedback on failure:

```
# Example of raw output from CI test run:
magic] node=0x1eaf2b0 resisptr=0x1db9e40
[magic] node=0x1ee0328 resisptr=0x1db7f50
[magic] Adding  collector; (Tnew = 1.53ps ; Tmin = 1.00ps)
[magic] Extracting emitter (Rnode = 27.00ohm ; Rthresh = 10.00ohm)
[magic] node=0x1c904e0 resisptr=0xc0000004
[magic] Aborted(Assertion failed: resptr->rn_ce != CLIENTDEFAULT, at: ResMerge.c,1126,ResCleanNode)
RuntimeError: Aborted(Assertion failed: resptr->rn_ce != CLIENTDEFAULT, at: ResMerge.c,1126,ResCleanNode)
RuntimeError: Aborted(Assertion failed: resptr->rn_ce != CLIENTDEFAULT, at: ResMerge.c,1126,ResCleanNode)
    at abort (file:///work/magic/npm/magic.js:474:41)
    at ___assert_fail (file:///work/magic/npm/magic.js:845:59)
    at magic.wasm.ResCleanNode (wasm://wasm/magic.wasm-02750336:wasm-function[1575]:0x186458)
    at magic.wasm.ResCleanUpEverything (wasm://wasm/magic.wasm-02750336:wasm-function[1491]:0x16e9cc)
    at magic.wasm.ResProcessNode (wasm://wasm/magic.wasm-02750336:wasm-function[1549]:0x17e9a7)
    at magic.wasm.ResCheckExtNodes (wasm://wasm/magic.wasm-02750336:wasm-function[1544]:0x17c1a2)
    at magic.wasm.ExtResisForDef (wasm://wasm/magic.wasm-02750336:wasm-function[1542]:0x17b947)
    at magic.wasm.CmdExtResis (wasm://wasm/magic.wasm-02750336:wasm-function[1546]:0x17dca3)
    at magic.wasm.WindExecute (wasm://wasm/magic.wasm-02750336:wasm-function[1425]:0x15d5fe)
    at magic.wasm.DBWcommands (wasm://wasm/magic.wasm-02750336:wasm-function[726]:0xaa203)

ext:   /work/magic/npm/examples/output/min.ext
Done.
```

This standard output above would ideally be augmented (either inline or as a copy below) with additional information to lookup original C file/line/snippet, this will save a lot of time.

Maybe this is something that can be knocked together as a python tool that can run in WASM CI and call the tooling and generate GitHub snippets.

The basic tools to run would be: 

```
# This is a tool available after EMSDK setup and source: source emsdk/emsdk_env.sh
$ which emsymbolizer
/work/magic/emsdk/upstream/emscripten/emsymbolizer
# We can take the info from the stacktrace and resolve to C file/line/byte
$ emsymbolizer /work/magic/npm/magic.wasm 0x186458
ResCleanNode
/work/magic/resis/ResMerge.c:1126:1
```

So for each stack line augment it before display like (one item displayed, but all can be augmented):

```
    at magic.wasm.ResCleanNode (wasm://wasm/magic.wasm-02750336:wasm-function[1575]:0x186458) => ResCleanNode /work/magic/resis/ResMerge.c:1126:1
```

Then for each valid entry account turn into a GitHub comment snippet, maybe this needs to be displayed as GITHUB_STEP_SUMMARY information that is only generated when necessary (from the GITHUB_SHA the relative project path, line number `#L1126`:

https://github.com/RTimothyEdwards/magic/blob/43e4cf9b038772c908eb57112a88afc0f91859ea/resis/ResMerge.c#L1126

or expand by 1 line either side `#L1125-L1127`:

https://github.com/RTimothyEdwards/magic/blob/43e4cf9b038772c908eb57112a88afc0f91859ea/resis/ResMerge.c#L1125-L1127

`https://github.com/RTimothyEdwards/magic/blob/43e4cf9b038772c908eb57112a88afc0f91859ea/resis/ResMerge.c#L1125-L1127`

Then have the python tool just run on the stdout/stderr output to transform the stack traces and emitted $GITHUB_STEP_SUMMARY info when something is found.  This will speed up resolving issues WASM detects by transforming them into C line level information directly in CI.

GitHub document link on line highlighting https://github.com/github/docs/blob/main/content/get-started/writing-on-github/working-with-advanced-formatting/creating-a-permanent-link-to-a-code-snippet.md

npm/build.sh
Intubun: Good catch - fixed it. Wrapped the test step in a subshell, so cd cant leak into --pack

npm/examples/extract-tcl.js
Intubun: Done. The example runners were calling console.error(e.message ?? e), which dropped e.stack — the part with the wasm-function[...]:0x... offsets. Added a
shared reportError() in helpers.js/helpers-tcl.js that prints the full stack:

export function reportError(e) {
  console.error(e?.stack ?? String(e?.message ?? e));
}

runScript() now wraps each command so the failing step is named before the error propagates:

try {
  this._run(l);
} catch (e) {
  console.error(`[magic] command failed: ${l}`);
  throw e;
}

All standalone runners and the suites use it, so node examples/extract.js now shows the full trace instead of a one-liner. And all.js/all-tcl.js already isolate each test and continue on failure, printing an N/M passed summary, so one failure doesn't stop the rest.

Verified on the failing extract test:
[magic] command failed: extresist all
FAIL  memory access out of bounds
RuntimeError: memory access out of bounds
    at magic.wasm.ResSimplifyNet  (wasm-function[1504]:0x13ae6b)
    at magic.wasm.ResDoSimplify   (wasm-function[1513]:0x13d2ca)
    at magic.wasm.ResProcessNode  ...
…with gds/drc/cif still running (3/4 passed).

RTimothyEdwards (approved):
After carefully looking through everything, and after having corrected the issue from my commit of a few days ago that caused the CI to fail, I approve this and will go ahead and merge it.

RTimothyEdwards:
Pulled and merged to the repository at opencircuitdesign.com.  The github mirror will update overnight.

PR #521, opened 2026-05-25
commands: Fix parsing of plane properties in case of single string

### LINKED ISSUE #520: Setting MASKHINTS_PPLUS either doesn't work or takes a while, non-deterministic
In the attached reproducer, I flatten a cell to generate a GDS and set MASKHINTS_PPLUS

If you run `./run.sh` several time, some times it will go fast and the property `MASKHINTS_PPLUS` isn't properly set ( I save debug `.mag`  at the end of the script and you can see it's empty ). Or it will take time ( like up to 30 sec ! ) but then the property is set properly ...

Neither is really good ... ideally just setting a property should both be quick and work 😅 

[issue.zip](https://github.com/user-attachments/files/28215663/issue.zip)

smunaut:
The previous code would iterate over `proplen` chunks of strings but would not initialize `proplen` for plane properties.

Technically for plane properties you don't need to "pre-count" the string chunks but given how the code is currently written it's easier to do the counting in all cases.

Also makes sure proprec is init to NULL so that if argument is empty, property is set to NULL and not random value on the stack.

Fixes #520

### DIFF
--- a/commands/CmdLQ.c
+++ b/commands/CmdLQ.c
@@ -2324,7 +2324,7 @@ CmdDoProperty(
     TxCommand *cmd,
     int argstart)
 {
-    PropertyRecord *proprec;
+    PropertyRecord *proprec = NULL;
     char *value;
     bool propfound, dolist;
     int proptype, proplen, propvalue, i;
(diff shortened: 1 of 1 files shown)


RTimothyEdwards:
I like issues that come with solutions!

RTimothyEdwards (approved):
Looks good to me, thanks.

RTimothyEdwards:
Pulled and merged to opencircuitdesign.com (magic version 8.3.650).

PR #526, opened 2026-06-09
Convert K&R function definitions to ANSI C; fix GCC 16 / C23 build

gonsolo:
Converts the old-style (K&R) function definitions across the tree to ANSI C
prototypes, formatted in the project's convention — the return type on its own
line, and each parameter on its own line indented four spaces with its original
comment retained — and adds the `-std=gnu17` flag needed for the GCC 16 / C23
build.

The change is organized as **one commit per directory** on top of a single
`build` commit, in dependency order (low-level first). Each directory commit
bundles its own header changes together with its definitions, so it is
self-consistent and every prefix of the series builds; it can be reviewed and
merged a directory at a time.

- All ~1162 K&R definitions converted (complete).
- Explicit prototypes / forward typedefs added where the now-stricter prototypes
  require them (`FontChar`, `TileType`, the netlist types); generic callbacks
  (e.g. `TechAddClient`) kept unprototyped; callback casts added where typed
  functions are stored in generic `(*)()` slots.
- One latent bug fixed: `dbStampFunc` was defined with two parameters but called
  recursively with one (hidden by the old K&R declaration).

Verified with a clean-from-scratch GCC 16 build: 0 errors, only the same benign
`-Wdiscarded-qualifiers` warnings as before; `tclmagic.so` links.


dlmiles:
AI generated, yes this could have been done this way a while ago using such a method.

Please take a look at the commit history for similar work done to the standard we are looking for.

This does not follow a number of already established source conventions in the project.

FWIW I do have the over 10 subdirs already converted, as we are finding out with AI, the time sink is in QA and testing (not creating/modifying code) and the only reason it has not been released sooner is that it hasn't passed my own standards for QA for release.


Maybe specific compiler (GCC16) concerns can be fixed with `CFLAGS="-std=c17 -g -D_DEFAULT_SOURCE=1" ./configure`

gonsolo:
> Please take a look at the commit history for similar work done to the standard we are looking for.

This uses c17 to keep the changes small (ahem). Later standards can be done subsequently.

> This does not follow a number of already established source conventions in the project.

Ok.

> FWIW I do have the over 10 subdirs already converted, as we are finding out with AI, the time sink is in QA and testing (not creating/modifying code) and the only reason it has not been released sooner is that it hasn't passed my own standards for QA for release.

I told Claude to write a smoke test (it's in the commit message). I was looking for a testsuite but didn't find one, that's the real QA in my opinion. Claude could be told to emit one for every line he changed and test it before and after the conversion. That and two or more humans looking over the code and testsuite would be a real improvement.

RTimothyEdwards:
@gonsolo :  As far as "established source conventions", please look at, for example, something like cif/CIFwrite.c line 118.  The question is not one of syntax but of style:
```
bool
CIFWrite(
    CellDef *rootDef,     /* Pointer to CellDef to be written */
    FILE *f)                       /* Open output file */
{
    ....
```
Each argument is on a separate line, indented four spaces.  An explanation of each argument (if available) follows on the same line.  The return value type is on a separate line above the routine name.  Arguments which are unused are so stated in the comment.

If the style is followed, then I am okay with reviewing and approving fixes in batches of up to a few hundred at a time.  Darryl has been feeding me fixes one directory at a time, which I think is a good practice.

gonsolo:
@RTimothyEdwards thanks for the style guidance — I've reworked the PR to follow it.

- **Style**: every converted definition now matches the convention from `cif/CIFwrite.c` — return type on its own line, each argument on its own line indented four spaces, with the original per-argument comments retained. Function-pointer and multi-argument parameters are handled.
- **Split per directory**: it's now one commit per directory (`utils`, `database`, `graphics`, … `windows`, `commands`, `oa`), preceded by a single `build` commit for the `-std=gnu17` flag. Each directory commit bundles its header changes with its definitions, so every prefix builds and you can review/merge a directory at a time.
- **Completeness**: the previous revision converted only ~378 of ~1162 K&R definitions (it had skipped those with multi-line argument comments, function-pointer parameters, or multi-line signatures). All of them are converted now.
- **Semantic fixes** the stricter prototypes surfaced: explicit prototypes where `bool`/`float`/`char` arguments clashed with the old `()` form; forward typedefs for `FontChar`, `TileType`, and the netlist types; `TechAddClient` kept generic (its per-section callbacks vary) with `opt` typed `int`; callback casts where typed functions go into generic `(*)()` slots; and one latent bug fix — `dbStampFunc` was defined with two parameters but called recursively with one.

Verified with a clean-from-scratch GCC 16 build: 0 errors (only the same benign `-Wdiscarded-qualifiers` warnings as before), and `tclmagic.so` links.

If the format looks right on, say, the `utils` or `database` commit, I'm glad to adjust anything before you go through the rest.

PR #527, opened 2026-06-10
Accept release tags without v prefix in the WASM npm publish gate.

Intubun:
The AppImage release workflows trigger on any tag (the project tags releases as e.g. 8.3.662, without a v prefix), but the npm publish gate in main-wasm.yml required a v-prefixed tag, so a normal release tag built the AppImages but never published the npm package to GitHub Packages.  The gate now accepts both forms, so a single tag releases everything.

### DIFF
--- a/.github/workflows/main-wasm.yml
+++ b/.github/workflows/main-wasm.yml
@@ -2,15 +2,15 @@ name: CI-wasm
 
 # Builds the Magic WebAssembly target (both the non-TCL and TCL variants)
 # on every push and pull request as a CI check. **Publishing** only happens
-# when a release tag of the form v<x.y.z>... is pushed — that gate is the
-# manual release trigger:
+# when a release tag of the form <x.y.z>... (optionally v-prefixed) is
+# pushed — the same tag that triggers the AppImage release workflows:
 #
 #   # bump magic/VERSION, commit, push to default branch
-#   git tag v8.3.638
-#   git push origin v8.3.638
+#   git tag 8.3.638
+#   git push origin 8.3.638
 #
-# The tag name (minus the leading "v") provides the base; the workflow appends
-# the commit date and short SHA: v8.3.799 → 8.3.799020261231+git01234cde.
+# The tag name (minus any leading "v") provides the base; the workflow appends
+# the commit date and short SHA: 8.3.799 → 8.3.799020261231+git01234cde.
 # Forks publish under their own namespace via the @<owner>/ scope.
 
 on:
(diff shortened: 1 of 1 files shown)


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged, and the version updated, so the behavior with respect to the version change can be checked when the github mirror updates tonight.

dlmiles:
@RTimothyEdwards  @Intubun 

Are people ok with me deleting (if GitHub lets me are there a usage/time limits on that action) the older packages at: https://github.com/RTimothyEdwards/magic/pkgs/npm/magic-vlsi-wasm

So the only version is the one just built, as kind of the first initial version ? 

Also it would be ideal if the GitHub artifact action:
 * Has one line additional header line (in the GITHUB_STEP_SUMMARY table) of the generated version string the action relates to, https://github.com/RTimothyEdwards/magic/actions/runs/27329734625  maybe `Package Version: @RTimothyEdwards/magic-vlsi-wasm@8.3.663020260610`  should this have +git suffix as well ?  or maybe it is in the package.json as metadata.
 * Also the download file artifact has the version string in the filename
 Is that something @Intubun can action in a newPR ?  maybe `magic-vlsi-wasm-npm-8.3.663020260610+git0123cdef`
 
 Other than this everything is looking good from my side, will download later artifact and try to be a consumer.

---

UPDATE

```
$ unzip -lv magic-vlsi-wasm-npm.zip 
Archive:  magic-vlsi-wasm-npm.zip
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
 6525919  Defl:N  6463561   1% 06-11-2026 07:05 c299af23  RTimothyEdwards-magic-vlsi-wasm-8.3.663020260610.tgz
--------          -------  ---                            -------
 6525919          6463561   1%                            1 file
```

An archive containing an archive is not ideal, can the filename be: `RTimothyEdwards-magic-vlsi-wasm-8.3.663020260610+git0123cdef`  note I added back on the GIT suffix part "+git0123cdef" because the CI might work on non-tagged versions, like in a developer clone/fork in GitHub so it keeps the unique filename.

I guess a misunderstanding how the GitHub Action upload aritfacts can work, as the only way to get the files is via an archive (like ZIP) as one complete unit, then input can be a collection of files.

---

Also are the *.map being copied (and other debug information, when available), it maybe that it is not being generated, might need `-gsource-map` option to build ?

---

UPDATE2:

`TCL ref 	core-9-0-b3`  maybe we should filter out the alpha/beta versions ?  I think core-9-0-3 is the one it should build against right now, so it selected the beta3 version instead of the 9-0-3 release version.  Also note they have `core-9.0.3-rc1` so removal of any `*-rc[0-9]*` and filter for `*-[0-9]+` should aim for the release version of Tcl.

RTimothyEdwards:
I'm okay with you deleting older packages.  I'll let Enno speak for himself.

Intubun:
yeah older packages are really only used by myself. I have to update them im my projects, but its manageable

dlmiles:
@Intubun are you actively working on these packaging / support improvements from my comment above

Intubun:
not currently. But ill work on it once I find the time for it.

dlmiles:
Most of these items were resolve with #543 
* generate and package *.map debug info
* don't build against alpha/beta/rc TkTcl versions
* don't tgz inside artifact zip
* STEP SUMMARY contains version info with git id


I did try deleting the unwanted NPM releases (when working through the long-term versioning concerns) but there is no control to me to delete, it was only a cosmetic thing to cleanup after making a mess working through initial CI issues.

PR #531, opened 2026-06-29
codebase-wide: use flexible arrays where appropriate

donn:
In platforms with bounds-checking enabled, placeholder values for array sizes in structs that may be exceeded always result in a sigtrap.

This patch updates the following classes to use flexible arrays at the end of structs:
- database
  - Label
  - PropertyRecord (union of three flexible arrays and a pointer*)
  - cellUE
- extflat
  - HierName
  - EFAttr
  - Dev
- utils
  - HashEntry (union of two flexible arrays and a pointer*)
   - internalUndoEvent: Was a plain int, replaced with a flexible char array

Additionally, the database struct editUE, which would have just a single flexible-array member, was removed and simply replaced with a character pointer.

Where possible, allocation size macros have been introduced.

---

* It is noted flexible-length arrays as part of unions are not part of the C language spec and are a GNU99 extension, however, it is supported by both GCC and Clang.

---

I tested the LibreLane Magic unit test set and it came out fine but any further testing is appreciated. Especially undoing in the GUI, because that was strangely one of the places where the semi-flexible array sizes were used.

### DIFF
--- a/calma/CalmaRdcl.c
+++ b/calma/CalmaRdcl.c
@@ -513,7 +513,7 @@ calmaParseStructure(
 	/* the same way that cell references are handled in .mag files.	*/
 	DBPathSubstitute(filename, cstring, cifReadCellDef);
 
-	proprec = (PropertyRecord *)mallocMagic(sizeof(PropertyRecord));
+	proprec = (PropertyRecord *)mallocMagic(dlongPropertyRecordSize(1));
 	proprec->prop_type = PROPERTY_TYPE_DOUBLE;
 	proprec->prop_len = 1;
 	proprec->prop_value.prop_double[0] = filepos;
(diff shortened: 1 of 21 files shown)


donn:
In a fit of irony, I broke it on Linux. Let me see what gcc flags I can make use of here.

EDIT: This is what I meant…

```
../database/database.h:738:15: error: flexible array member in union
  738 |         int   prop_integer[];   /* For PROPERTY_TYPE_INTEGER or _DIMENSION */
```

donn:
Flexible arrays inside unions are only supported in GCC 15 or newer… Great…

RTimothyEdwards:
I think that would have been  more elegantly handled by splitting the `PropertyRecord` into four different structures for string, integer, double, and plane rather than trying to deal with the union.  Most of it would still be the same but it would get rid of the `FLEX_UNION_MEMBER` awkwardness.  But then it would still exist for `HashEntry`.

Any issue with `prop_len` would have been something introduced by me, recently, and may well just be incorrect.

RTimothyEdwards:
Pulled and merged to opencircuitdesign.com.

Issue #533, opened 2026-07-02
An update broke the wasm build in an obscure way


RTimothyEdwards:
@Intubun : 

The last push caused an error with the wasm build, although I'm not at all clear why anything that was changed should be an issue for wasm.

Specifically, the Tcl build of wasm can't find the header file "sys/queue.h".

The Google AI says:

"This error occurs because the WebAssembly C/C++ runtime (Emscripten) relies on the musl standard C library, which natively lacks the BSD-specific <sys/queue.h> header file required by tclEpollNotfy.c. 
To fix this, you can provide the missing header from the BSD compatibility libraries using Emscripten’s -sUSE_BETA_PORTS or by manually downloading a standalone version."

Can you confirm the solution and tell me where it needs to be applied?

Intubun:
Yeah, the WASM enviroment lacks any sort of specific OS Calls. So they need to be abstracted away though compatibility libs. Ill look at it probably Saturday evening

dlmiles:
https://github.com/RTimothyEdwards/magic/actions/runs/28719342173

emscripten latest = 6.0.2 (which has broken the build)
emscripten 6.0.1 still works

@Intubun thanks for putting in the changes requested during the original PR as it makes this diagnosis just a 4 mins turn around.  Reminder if you have free capacity to work through the matters arising from https://github.com/RTimothyEdwards/magic/pull/527#issuecomment-4680652433 so the WASM build can be supported

If this persists to be a problem (due to 6.0.3 keeping this change and it not being an emscripten release oversight):

* inhibit magic WASM autoconf config.h HAVE_SYS_EPOLL_H / NOTIFIER_EPOLL support from the WASM Tcl build so it falls back to select(), like Tcl8 uses.
OR
* confirm `<sys/event.h>` is permanently removed from emscripten (and not an error), then see what breaks (in Tcl build) without it defined, confirm the header is actually needed, if Tcl still builds propose a patch to Tcl 9.x to autoconf the existence of the header file maybe as HAVE_SYS_EVENT_H.

RTimothyEdwards:
I have just merged the pull request that should fix this issue;  if the github mirror copy tonight doesn't result in any errors, then I will close this issue tomorrow.

Intubun:
It shouldnt. Already passed all the tests on my repo.

RTimothyEdwards:
Looks like it all went smoothly, so I'll go ahead and close the issue.

PR #536, opened 2026-07-07
Bugfix for WASM Build with emscripten >6.0.2

Intubun:
This PR is related to Issue #533 . Here a small summary:

- What broke it — emscripten 6.0.2 (6.0.1 was fine) made two sysroot/default changes: it started shipping a
  <sys/epoll.h> stub, and it dropped wasmBinary from the default INCOMING_MODULE_JS_API list. Both broke the
  WASM build in different ways.
  - toolchains/emscripten/build-tcl-wasm.sh — Added ac_cv_header_sys_epoll_h=no to TCL's configure
  invocation, forcing the select()-based notifier. The new <sys/epoll.h> stub made TCL pick the epoll
  notifier (tclEpollNotfy.c), which needs <sys/queue.h> — a header emscripten doesn't provide — causing a
  compile-time fatal error: 'sys/queue.h' file not found.
  - toolchains/emscripten/defs.mak — Added -sINCOMING_MODULE_JS_API= (emscripten's full default list plus
  wasmBinary). Because wasmBinary was removed from the default, the loaders passing Module.wasmBinary
  triggered a hard runtime abort under -sASSERTIONS=1 (Module.wasmBinary … not included in
  INCOMING_MODULE_JS_API), failing every example test.

### DIFF
--- a/toolchains/emscripten/build-tcl-wasm.sh
+++ b/toolchains/emscripten/build-tcl-wasm.sh
@@ -93,6 +93,18 @@ cd "$OUT"
 # --disable-shared      we statically link libtcl into magic.wasm.
 # --disable-load        no dynamic loading inside wasm.
 # --enable-symbols=no   release (-O2). Override CFLAGS to add -g for debug.
+#
+# ac_cv_header_sys_epoll_h=no forces TCL onto its select()-based notifier.
+# On a Linux host, TCL's configure runs `AC_CHECK_HEADERS([sys/epoll.h])` and,
+# if found, defines NOTIFIER_EPOLL and compiles tclEpollNotfy.c. Emscripten
+# 6.0.2 started shipping a <sys/epoll.h> stub in its sysroot (6.0.1 did not),
+# so that probe now passes — but tclEpollNotfy.c also needs <sys/queue.h>,
+# which emscripten does NOT provide, so the build dies with
+#   tclEpollNotfy.c: fatal error: 'sys/queue.h' file not found
+# The select() notifier is the correct choice for single-threaded WASM anyway
+# (emscripten's epoll is a stub), so we pin it here rather than track the
+# emscripten sysroot. Overriding the autoconf cache var is cleaner than
+# patching the read-only TCL source tree.
 if [ ! -f Makefile ]; then
   echo "=== emconfigure ==="
   CFLAGS="-O2 -sUSE_ZLIB=1" \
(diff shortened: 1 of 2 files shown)


Intubun:
@dlmiles Or would it be better to have automatically latest flow, and when it fails, that it falls back to the last known good version to check against it and it then just gives a warning, if it builds with the older emscripten version?

dlmiles:
This is fine, from your comment it looks like `<sys/epoll.h>` is new/recent support in WASM so we can expect more breakage in that area until the bugs are ironed out (by enscripten), so we sidestep that by using `<sys/select.h>` which is what it always did before.


My preference is to to persevere with use of `latest` for a while, just to canvas how much and how often breakage occurs.

It is difficult for a CI to know what a "last known good version" is, that would still degenerate into needing to be a manually maintained thing (like workflow environment variable updated from time to time).  But doing such a fallback could make the NPM release sporadic, if we assumed the fallback idea works and we always get a release either the latest or the fallback version. You get a situation to the customer: one release works, next release doesn't, next release does, next release doesn't - the releases that do work are the fallback releases, and the release that didn't work are because it successfully built against latest but there is some other bug that persists in the upstream project that we only see when a build against latest is successful.  Such a situation is not good for our customer project perception - as the releases now become unpredictable.  Better to always move forward when you can and never backwards, so it stays broken until reported and then fixed.


In the future if this kind of breakage gets too much maintenance (too brittle), it would not be a problem with having a `ci-wasm.yml` and a `canary-wasm.yml`:

* canary-wasm.yml a copy of the current workflow, always setup to the latest stable release (just as now) but never publishing (to npm) and only runs on new tags (not every PR).  Like the other canary-\*.yml  Designed to provide early warning and visibility of the tcl/emscripten projects at the (stable) bleeding-edge and provide integration feedback.  It is ok for canary-\* workflows to fail because of this.

* ci-wasm.yml a copy of the current workflow, continues to run on every PR and tag (just as now), setup to locked versions of known to be good upstream project (and we have/accept periodic maintenance to bump the versions forward from time to time ~6 months), publish on tag pushed (just as now), designed to produce reliable generation of NPM every release lagging the bleeding-edge.  The every 6 months maintenance becomes: you use the feedback from the canary-wasm.yml to make the version bump decision.

RTimothyEdwards (approved):
Looks good to me.  Based on the discussion, I will go ahead and approve and merge this, which should solve the immediate CI failure.  Any adjustments having to do with fallback versions and such can be added later in a separate pull request, if deemed necessary.

RTimothyEdwards:
Pulled and merged to opencircuitdesign.com.

PR #540, opened 2026-07-08
oa: subdir module

dlmiles:
The purpose is to make it buildable under CI so it does not bitrot over time.


RTimothyEdwards (approved):
Thanks.  I think my preferred solution would be to remove the directory altogether, given that the feature was never completely developed.  However, this is better than leaving it there and, as you said, collecting bit-rot.

RTimothyEdwards:
Pulled and merged to the repository at opencircuitdesign.com

PR #544, opened 2026-07-22
lisp: subdir module

dlmiles:
The purpose is to make it buildable under CI so it does not bitrot over time.


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:

PR #545, opened 2026-07-22
crosshair: prevent crash from command usage in headless mode

dlmiles:


RTimothyEdwards (approved):
Good catch;  didn't know this failed in headless mode.

RTimothyEdwards:

